PR #1, opened 2018-07-11
Initial driver push

msau42:


deploy/kubernetes/manifests/controller.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
+              value: /csi/csi.sock
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+        - name: gcp-filestore-driver
+          imagePullPolicy: Always
+          image: gcr.io/google-containers/volume-csi/gcp-filestore-csi-driver:v0.1.0
```
davidz627: do these exist yet?

deploy/kubernetes/manifests/node.yaml
```diff
@@ -0,0 +1,79 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: kubelet-dir
+              mountPath: /var/lib/kubelet
+              mountPropagation: "Bidirectional"
+            - name: plugin-dir
+              mountPath: /csi
+        - name: nfs-services
```
davidz627: ? can this be done in the same container based on `--node=true` maybe?

davidz627:
besides those things
/lgtm

deploy/kubernetes/manifests/controller.yaml
msau42: Nope, I'm going to tag and push it soon

deploy/kubernetes/manifests/node.yaml
msau42: I wanted to keep it separate because:
1) it launches a bunch of other processes
2) it can have its own health monitoring/probes
3) it's not related to csi, it's a system dependency

PR #7, opened 2018-07-14
Wait for operation in cloud provider

### LINKED ISSUE #5: DeleteVolume isn't actually deleting the filestore instance
/kind bug

msau42:
The Kubernetes external-provisioner timeout is currently too short for a create/delete operation to succeed in one call.  However, non-Kubernetes implementations may not have such a short timeout, so the cloud provider calls should try to wait for the operation to complete if able to.

Fixes #5

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -19,11 +19,14 @@ package file
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"google.golang.org/api/googleapi"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	beta "sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/generated/file/v1beta1"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
@@ -62,8 +65,9 @@ type gcfsServiceManager struct {
 }
 
 const (
-	locationURIFmt = "projects/%s/locations/%s"
-	instanceURIFmt = locationURIFmt + "/instances/%s"
+	locationURIFmt  = "projects/%s/locations/%s"
+	instanceURIFmt  = locationURIFmt + "/instances/%s"
+	operationURIFmt = locationURIFmt + "/operations/%s"
 )
 
 var _ Service = &gcfsServiceManager{}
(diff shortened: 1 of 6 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -206,6 +259,21 @@ func instanceURI(project, location, name string) string {
 	return fmt.Sprintf(instanceURIFmt, project, location, name)
 }
 
+func operationURI(project, location, name string) string {
+	return fmt.Sprintf(operationURIFmt, project, location, name)
+}
+
+func getInstanceNameFromURI(uri string) (project, location, name string, err error) {
```
msau42: this could use some unit tests

pkg/cloud_provider/file/file.go
```diff
@@ -141,10 +155,16 @@ func (manager *gcfsServiceManager) GetInstance(ctx context.Context, obj *Instanc
 	return nil, nil
 }
 
-func cloudInstanceToServiceInstance(instance *beta.Instance) *Instance {
+func cloudInstanceToServiceInstance(instance *beta.Instance) (*Instance, error) {
```
davidz627: I would rename `Instance` to `ServiceInstance` or something to disambiguate from the cloud instance. Currently its hard to tell when you have which one

pkg/cloud_provider/file/file.go
```diff
@@ -189,13 +209,46 @@ func (manager *gcfsServiceManager) DeleteInstance(ctx context.Context, obj *Inst
...
 	}
 
-	// Always return error and check for instance in the subsequent calls
-	return fmt.Errorf("DeleteInstance operation started")
+	err = manager.waitForOp(ctx, op)
+	if err != nil {
+		return fmt.Errorf("WaitFor DeleteInstance operation failed: %v", err)
+	}
+
+	instance, err = manager.GetInstance(ctx, obj)
```
davidz627: not sure I like that `!isNotFoundErr(err)` is hidden inside `GetInstance` and it simply doesn't return an instance instead of throwing an error. Technically it might be better but it also goes against the pattern that we see in all of Kubernetes so to me it was fairly confusing

pkg/cloud_provider/file/file.go
davidz627: agreed

pkg/cloud_provider/file/file.go
msau42: done

pkg/cloud_provider/file/file.go
msau42: done

pkg/cloud_provider/file/file.go
msau42: done

PR #13, opened 2018-08-02
Update vendor to get csi-sanity updates

msau42:
csi-sanity issues have been fixed so now the tests can pass and be enabled


davidz627:
I would recommend locking your dep to a specific version instead of master. See:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/Gopkg.toml#L79

msau42:
I need the latest version to get a fix, so I've locked it to the latest

davidz627:
/lgtm

@msau42 should the fix be cherrypicked to the v0.3.0 release of the csi-test driver then? I see no reason why we should be using an intermediate version for a critical fix. Could you open an issue for the cherry-pick if required?

PR #16, opened 2018-08-07
Adding template for E2E tests

krunaljain:
Adding Single and Multi Zone E2E tests


msau42:
/ok-to-test

You will also need to vendor in the dependencies.  Run "dep --ensure" and checkin the vendor updates as a 2nd commit

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -0,0 +1,178 @@
...
+	readyState             = "READY"
+	standardDiskType       = "pd-standard"
+	ssdDiskType            = "pd-ssd"
+)
+
+var _ = Describe("GCFS CSI Driver", func() {
+
+	It("Should create->attach->stage->mount volume and check if it is writable, then unmount->unstage->detach->delete and check disk is deleted", func() {
+		Expect(testInstances).NotTo(BeEmpty())
+		testContext, err := testutils.GCEClientAndDriverSetup(testInstances[0])
```
msau42: Is the purpose to run the GCE PD driver tests or to run the GCFS driver?

msau42: If the purpose is to run the GCFS driver, then try incorporating the test I wrote using the old framework here: https://github.com/msau42/gcp-filestore-csi-driver/blob/e2e-test/test/e2e/filestore_test.go

msau42:
also needs the Gopkg files

PR #23, opened 2018-08-24
rename e2e scripts

msau42:


davidz627:
/lgtm

Is there a reason why we prefer underline to hyphen? (asking so I can decide whether to change all of mine as well)

msau42:
Mainly personal preference.  In kuberentes, it seems to be inconsistent.  Most of our volume files use underscore, but most of the hack scripts use hyphen

PR #36, opened 2019-07-23
Windows SMB Support

almos98:
Refactored some Linux specific code to allow requests to be made when running on Windows to mount SMB shares.

The code has been manualy tested end-to-end and successfully mounts an SMB share on a Windows node. However, it doesn't run containerized yet (see this [enhancement](https://github.com/kubernetes/enhancements/issues/1122)) and hasn't been tested against GCP FileStore because SMB support is not available.


cmd/main.go
```diff
@@ -32,6 +32,8 @@ var (
 	nodeID        = flag.String("nodeid", "", "node id")
 	runController = flag.Bool("controller", false, "run controller service")
 	runNode       = flag.Bool("node", false, "run node service")
+	username      = flag.String("username", "", "Windows username")
```
msau42: Are these credentials for the samba server?

pkg/csi_driver/node.go
```diff
@@ -67,7 +68,7 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 	// Check if mount already exists
 	mounted, err := s.isDirMounted(targetPath)
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
msau42: maybe have `isDirMounted` return `false, nil` if the directory doesn't exist

pkg/csi_driver/node.go
```diff
@@ -79,16 +80,31 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+	// FileSystem type
+	fstype := "nfs"
+
 	// Mount options
 	options := []string{}
+
+	// Windows specific values
+	if runtime.GOOS == "windows" {
+		source = fmt.Sprintf("\\\\%s\\%s", attr[attrIp], attr[attrVolume])
+		fstype = "cifs"
```
msau42: should we validate fstype in `validateVolumeCapabilities`?

pkg/csi_driver/node.go
```diff
@@ -79,16 +80,31 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+
+		// Login credentials
+
+		options = append(options, s.driver.config.Username)
```
msau42: we should have these credentials passed in through the NodePublish secret, which gets set in `NodePublishVolumeRequest.secrets`. https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#node-publish-secret

msau42: Do we need these secrets on unmount too?

pkg/csi_driver/node.go
almos98: The [mount utility](https://github.com/kubernetes/kubernetes/blob/5b496fe8f5933de9b53e35780a486c8f7736d02b/pkg/util/mount/mount_windows.go#L85) for Windows only supports CIFS so I think it's okay to leave this hardcoded for now.

cmd/main.go
almos98: I haven't tested this for Samba servers, but they're the credentials for an SMB share on Windows.

pkg/csi_driver/node.go
almos98: If `isDirMounted` returns `false, nil` if the directory doesn't exist there is code in [`unmountPath`](https://github.com/almos98/gcp-filestore-csi-driver/blob/83a40d3799e620f682bb3dca054f139359877f80/pkg/csi_driver/node.go#L179) that might not be reached. What should I do about that code?

pkg/csi_driver/node.go
almos98: I'm not sure how I would implement passing through NodePublish secret because [mount_windows.go](https://github.com/kubernetes/kubernetes/blob/5b496fe8f5933de9b53e35780a486c8f7736d02b/pkg/util/mount/mount_windows.go#L94) requires the username and password to be in options.

Secrets aren't needed to unmount.

pkg/csi_driver/node.go
msau42: I think the unmountPath logic would work fine.  If the directory doesn't exist, `unmountPath` returns success. We would just lose the log message, which maybe could be added into `isDirMounted` if we wanted to keep it.

pkg/csi_driver/node.go
msau42: Correct, but the fstype is set by the user and passed to us.  So the point of `validateVolumeCapabilities` is to make sure we're able to support what the user requested.

pkg/csi_driver/node.go
msau42: That's fine, we can take the secrets from `NodePublishVolumeRequest` and use them to construct the `options` passed into the mount call.  This way, we can support per-volume credentials.

cmd/main.go
davidz627: wonder if this is more of a runtime configuration that may be different per volume/namespace/storage class? If so, might be better to provide through secrets instead of a flag on the binary

pkg/csi_driver/node.go
```diff
@@ -79,16 +80,31 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
 	source := fmt.Sprintf("%s:/%s", attr[attrIp], attr[attrVolume])
 
+	// FileSystem type
+	fstype := "nfs"
+
 	// Mount options
 	options := []string{}
+
+	// Windows specific values
+	if runtime.GOOS == "windows" {
```
davidz627: nit: create a util function `IsWindows` or something

cmd/main.go
almos98: I've implemented this using NodePublishSecrets but I'm not sure if I did it the right way. Feedback appreciated!

pkg/csi_driver/node.go
almos98: Currently don't keep the log message, let me know if you'd prefer to keep it.

pkg/csi_driver/node.go
almos98: This is how kubernetes/kubernetes checks for Windows in many [locations](https://github.com/kubernetes/kubernetes/search?q=GOOS+windows&unscoped_q=GOOS+windows). I don't think it's worth adding a wrapper.

pkg/csi_driver/node.go
```diff
@@ -79,16 +85,36 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	// Mount source
 	source := fmt.Sprintf("%s:/%s", attr[attrIp], attr[attrVolume])
 
+	// FileSystem type
```
yujuhong: Do we expect to support additional filesystem types in the future? (or, say, do we want to use this csi driver for all filesystem types supported by filestore?)

The pattern of hardcoding fstype according to the OS type seems a bit strange to me. Is this associated with any input and can be validated?

pkg/csi_driver/node.go
```diff
@@ -157,13 +183,24 @@ func validateVolumeAttributes(attr map[string]string) error {
 	return nil
 }
 
+func validateNodePublishSecrets(secrets map[string]string) error {
```
yujuhong: This is specific to SMB. Make that explicit in the function name.

Gopkg.toml
```diff
@@ -65,3 +65,7 @@
   go-tests = true
   unused-packages = true
   non-go = true
+
+[[override]]
+  name = "gopkg.in/fsnotify.v1"
+  source = "https://github.com/fsnotify/fsnotify.git"
```
yujuhong: missing new line at the end.

Gopkg.toml
```diff
@@ -65,3 +65,7 @@
   go-tests = true
   unused-packages = true
   non-go = true
+
+[[override]]
```
yujuhong: Why is "override" needed?

almos98: I'm not sure why, but when I was running `dep ensure -update` it would break. After looking around I found this [issue](https://github.com/kubernetes-sigs/kubebuilder/issues/316) so I applied the workaround I found there.

pkg/csi_driver/node.go
almos98: Before I made any changes the driver only supported "nfs" (it was hardcoded). [mount_windows.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/util/mount/mount_windows.go#L86) only supports CIFS.

I don't think it's associated with any input. Users don't get to choose the FS type in the case of FileStore.

pkg/csi_driver/node.go
almos98: Is `validateSmbNodePublishSecrets` okay for the name?

pkg/csi_driver/node.go
```diff
@@ -93,8 +98,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 
 		// Login credentials
 
-		options = append(options, s.driver.config.Username)
-		options = append(options, s.driver.config.Password)
+		secrets := req.GetNodePublishSecrets()
```
msau42: We should modify L52 above to make sure we don't log the contents of the secrets.  We may be able to use https://github.com/kubernetes-csi/csi-lib-utils/blob/master/protosanitizer/protosanitizer.go for this

pkg/csi_driver/node.go
```diff
@@ -93,8 +98,13 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+		if err := validateNodePublishSecrets(secrets); err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		options = append(options, secrets[smbUser])
+		options = append(options, secrets[smbPassword])
```
msau42: Can we enhance https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/node_test.go#L60 to verify that the mount options are correct?

pkg/csi_driver/node.go
almos98: Done.

pkg/csi_driver/node.go
almos98: Done.

pkg/csi_driver/node.go
almos98: Done?

pkg/csi_driver/node_test.go
```diff
@@ -177,11 +188,54 @@ func TestNodePublishVolume(t *testing.T) {
 			},
 			expectErr: true,
 		},
+		// TODO: enable this test when FakeMount supports Windows behavior
```
msau42: It might be cleaner to actually have a separate test function just for the windows cases.

pkg/csi_driver/node_test.go
```diff
@@ -106,27 +106,28 @@ func TestNodePublishVolume(t *testing.T) {
...
-						Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
-					},
-				},
-				VolumeAttributes: testVolumeAttributes,
-			},
-			actions: []mount.FakeAction{{Action: mount.FakeActionMount}},
-			// TODO: add all mount options into FakeMounter
-			expectedMount: &mount.MountPoint{Device: testDevice, Path: testTargetPath, Type: "nfs"},
-		},
+		// TODO: add all mount options into FakeMounter
```
msau42: why is this failing?

almos98: FakeMounter's behavior has changed after updating the vendor files. Before the update it would check each mount option and include it in the `expectedMount` Opts field. Now it just appends any options passed to `expectedMount` Opts field. The test expects no options but FakeMounter actually returns "foo" and "bar" as options.

msau42: I think the new behavior is correct.  The driver currently passes through all mount options to the mount command, so the expectedMount should have those mount options

almos98: Okay will change this test to reflect that then.

pkg/csi_driver/node_test.go
almos98: Done.

almos98:
@msau42 @yujuhong @jingxu97 I think this is ready for review now.
/cc adjackura

pjh:
cc @adjackura

pkg/csi_driver/node.go
```diff
@@ -68,8 +79,15 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	// Check if mount already exists
 	mounted, err := s.isDirMounted(targetPath)
 	if err != nil {
-		return nil, err
+		if goOs == "windows" {
+			if !os.IsNotExist(err) {
```
pjh: Add a comment here, something like "on Windows the mount target must not exist yet".

Also, you can combine with previous line: goOs == "windows"  and !os.IsNotExist(err)

pkg/csi_driver/node.go
pjh: @almos98 is this done now?

pkg/csi_driver/node_test.go
```diff
@@ -199,6 +206,92 @@ func TestNodePublishVolume(t *testing.T) {
...
+				VolumeAttributes:   testVolumeAttributes,
+				NodePublishSecrets: testWindowsSecrets,
+			},
+
+			actions:       []mount.FakeAction{{Action: mount.FakeActionMount}},
+			expectedMount: &mount.MountPoint{Device: testWindowsDevice, Path: testWindowsValidPath, Type: "cifs", Opts: []string{"foo", "bar"}},
+		},
+		{
+			osString: "windows",
+			name:     "windows no credentials",
```
pjh: For 100% coverage you could split this into two tests: one where smbUser is not set (but password is), one where smbPassword is not set (but user is). Both should return errors.

pjh:
LGTM!

@almos98 could you update the PR description to note that you've manually tested this end-to-end to confirm that it successfully mounts an SMB share on the Windows node, but 1) it doesn't run containerized yet (link to https://github.com/kubernetes/enhancements/issues/1122), and 2) it hasn't been tested against GCP Filestore (since SMB support is still not available). Thanks!

pkg/csi_driver/node.go
almos98: In the current code, if goOs is Windows and the error is IsNotExist, the code won't do anything. Example:
```go
if goOs == "windows" {
    if !os.IsNotExist(err) {
        return nil, err
    }
    // if goOs == "windows" and os.IsNotExist(err)
    //do nothing
} else {
    return nil, err
}
```

If L82 and L83 are combined then this is lost.

pkg/csi_driver/node_test.go
almos98: Done.

pkg/csi_driver/node.go
pjh: Oops, you're right. Looks good as it is.

msau42:
This is looking good.  Can you also do the following:
- squash your commits down into 2-3 commits. One for vendor updates, and 1 or 2 for the Windows implementation?  If you haven't done this before, create a duplicate branch first so you still have all your commits
- retitle the PR. this is more than a refactor
- did you use any scripts to manually test the csi driver?  If so, can you add them to test/experimental?  Or at least document the steps somewhere

msau42:
Also is it possible to enhance https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/test/run_unit.sh to also run the windows unit tests?  Maybe @yujuhong or @pjh can help with how to do that.

msau42:
> Also is it possible to enhance https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/test/run_unit.sh to also run the windows unit tests? Maybe Yu-Ju Hong or Peter Hornyack can help with how to do that.

Sorry I realized the unit tests were written in a way that it doesn't actually require a windows env to run.  So scratch that, I think it is fine the way it is.

almos98:
Squashed commits, changed PR title and documented testing steps!

pkg/csi_driver/node.go
```diff
@@ -157,6 +196,17 @@ func validateVolumeAttributes(attr map[string]string) error {
 	return nil
 }
 
+func validateSmbNodePublishSecrets(secrets map[string]string) error {
+	if secrets[smbUser] == "" {
```
msau42: does smb have any restrictions on what are valid values?

pkg/csi_driver/node_test.go
msau42: Is there something still preventing this test from being enabled?  If so, let's open an issue for it and reference it here

pkg/csi_driver/gcfs_driver.go
```diff
@@ -147,8 +147,7 @@ func (driver *GCFSDriver) validateVolumeCapability(c *csi.VolumeCapability) erro
 		return fmt.Errorf("driver only supports mount access type volume capability")
 	}
 	if mountType.FsType != "" {
-		// TODO: uncomment after https://github.com/kubernetes/kubernetes/issues/65122 is fixed
-		// return fmt.Errorf("driver does not support fstype %v", mountType.FsType)
+		return fmt.Errorf("driver does not support fstype %v", mountType.FsType)
```
msau42: actually can you revert this line.  Looks like the issue hasn't been completely fixed: https://github.com/kubernetes-csi/external-provisioner/issues/328

Makefile
```diff
@@ -27,6 +27,10 @@ local:
 	mkdir -p bin
 	go build -ldflags "-X main.vendorVersion=${VERSION}" -o bin/gcfs-csi-driver ./cmd/
 
+windows-local:
+	mkdir -p bin
+	GOOS=windows GOARCH=amd64 go build -ldflags "-X main.vendorVersion=${VERSION}" -o bin/gcfs-csi-driver.exe ./cmd/
+
 push:
 	docker push $(IMAGE):$(VERSION)
```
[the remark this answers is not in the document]
almos98: We will need to write a separate Dockerfile to build a Windows container. I haven't written one because the driver can't run from within a container on Windows. Can write one and include it if you'd like.

pkg/csi_driver/node.go
almos98: There's some restrictions but they're a bit convoluted and are caught during mount so I think checking they're not empty is good enough.
Example: the SMB username shouldn't be more than 20 characters long, not counting the domain name. Special characters, etc.

Passwords have no restriction.

Makefile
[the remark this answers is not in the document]
msau42: It may be good to add now so once CSI on windows is ready, we can easily leverage this

pkg/csi_driver/node_test.go
almos98: On Windows, if TargetPath exists it should return an error but FakeMount doesn't do that.

Makefile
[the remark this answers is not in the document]
almos98: Added a Dockerfile for windows under test/experimental but the driver has to be compiled first. I added a target in the Makefile to do this.

pjh:
@davidz627 @msau42 can you provide your final feedback and approval so that we can get this merged? Thanks!

pjh:
@almos98 consider squashing this back into two commits again.

Makefile
```diff
@@ -27,6 +27,15 @@ local:
 	mkdir -p bin
 	go build -ldflags "-X main.vendorVersion=${VERSION}" -o bin/gcfs-csi-driver ./cmd/
 
+windows: windows-local
+	cp Dockerfile Linux-Dockerfile
+	cp test/experimental/Dockerfile Dockerfile
+	docker build --build-arg TAG=$(VERSION) -t $(IMAGE)-windows:$(VERSION) .
```
msau42: We can use "-f" option to pass in a non-default Dockerfile.  That way we don't need to mess with the normal one.

almos98: Done.

pkg/csi_driver/node_test.go
almos98: Opened an issue on K/K: https://github.com/kubernetes/kubernetes/issues/81609

msau42:
/lgtm

Thanks!

PR #37, opened 2019-08-20
CSI Test Tool

almos98:
This PR adds a tool that allows the CSI driver to be exercised locally both on Linux and on Windows.

Two examples on how to use this tool:
NodePublishVolume:
```
.\csi-request.exe --request publish --host "windows-node-name" --volume "SMBShare" --targetPath "C:\Share" --username "smbuser" --password "foobar"
```

NodeUnpublishVolume:
```
.\csi-request.exe --request unpublish --targetPath “C:\Share”
```


hack/csi_test/cmd/main.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"flag"
+	"fmt"
+
+	csi "sigs.k8s.io/gcp-filestore-csi-driver/hack/csi_test/pkg/csi_tool"
+)
+
+var (
+	endpoint    = flag.String("endpoint", "/tmp/csi.sock", "CSI endpoint")
+	host        = flag.String("host", "localhost", "Host machine name")
```
msau42: should this be something more specific like "shareAddr"?

hack/csi_test/cmd/main.go
```diff
@@ -0,0 +1,40 @@
...
+	volume      = flag.String("volume", "", "Volume name")
```
msau42: would "shareName" make it more obvious? This is more inline with what Filestore calls it

hack/csi_test/pkg/csi_tool/client.go
```diff
@@ -0,0 +1,141 @@
...
+				AccessMode: &csi.VolumeCapability_AccessMode{
+					Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+				},
+			},
+			VolumeAttributes: map[string]string{
+				attrIP:     req.Host,
+				attrVolume: req.Volume,
+			},
+			NodePublishSecrets: map[string]string{
+				smbUser:     fmt.Sprintf("%s\\%s", req.Host, req.Username),
```
msau42: These should only be set if the args are non-nil

Gopkg.lock
```diff
@@ -406,9 +406,12 @@
   version = "v0.3.3"
 
 [[projects]]
-  digest = "1:bbe569477853f0bd2e49e8f214ab37981293683dff4a9815e95a1465c3918ee6"
+  digest = "1:aef2bfdbd53d02eb4f346775b9b78dbc15011092497a049bb5b433dfe3634847"
   name = "k8s.io/kubernetes"
-  packages = ["pkg/util/mount"]
+  packages = [
+    "pkg/util/mount",
+    "pkg/volume/csi/csiv0",
```
msau42: The csi spec should already be vendored in.  Do we really need to pull in the version from K8s?

almos98: Ah I forgot to change the path after moving the tool location. We don't need to pull it from K8s.

hack/csi_test/cmd/main.go
almos98: That makes more sense, will change it!

hack/csi_test/cmd/main.go
almos98: Also changing this, thanks!

pjh:
cc @adjackura

hack/csi_test/cmd/main.go
```diff
@@ -9,8 +9,8 @@ import (
 
 var (
 	endpoint    = flag.String("endpoint", "/tmp/csi.sock", "CSI endpoint")
-	host        = flag.String("host", "localhost", "Host machine name")
-	volume      = flag.String("volume", "", "Volume name")
+	shareAddr   = flag.String("shareAddr", "localhost", "Address of the share")
+	shareName   = flag.String("volume", "", "Volume name")
```
msau42: also update the flag name and description too?  Volume is ambiguous in the CSI context because it could mean the CSI volume name.

hack/csi_test/cmd/main.go
```diff
@@ -9,8 +9,8 @@ import (
 
 var (
```
msau42: Would it make sense to also have a help arg with more detailed examples of commands? Since it's not obvious from looking at the flags what's required for publish and what's required for unpublish

hack/csi_test/cmd/main.go
```diff
@@ -9,8 +9,8 @@ import (
...
 	endpoint    = flag.String("endpoint", "/tmp/csi.sock", "CSI endpoint")
-	host        = flag.String("host", "localhost", "Host machine name")
-	volume      = flag.String("volume", "", "Volume name")
+	shareAddr   = flag.String("shareAddr", "localhost", "Address of the share")
+	shareName   = flag.String("volume", "", "Volume name")
 	targetPath  = flag.String("targetPath", "", "Path to mount volume at")
 	request     = flag.String("request", "publish", "CSI request type")
```
msau42: make the options nodePublish and nodeUnpublish to align more with the CSI spec and disambiguate from controllerPublish/Unpublish

msau42: Also define them as consts since we use them in multiple places

hack/csi_test/cmd/main.go
almos98: Added a help message (it already supported -h flag), will this work?

hack/csi_request_tool/cmd/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+var (
+	endpoint    = flag.String("endpoint", "/tmp/csi.sock", "CSI endpoint")
+	shareAddr   = flag.String("shareAddr", "localhost", "Address of the share")
+	shareName   = flag.String("shareName", "", "Name of the share")
+	targetPath  = flag.String("targetPath", "", "Path to mount volume at")
+	smbUsername = flag.String("smbUsername", "", "WINDOWS ONLY: Username to login to SMB share")
+	smbPassword = flag.String("smbPassword", "", "WINDOWS ONLY: Password to login to SMB share")
+
+	request = flag.String("request", "", requestHelpMessage)
```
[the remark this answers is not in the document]
almos98: I implemented something using this. Could you PTAL?

hack/csi_client/cmd/main.go
```diff
@@ -1,34 +1,45 @@
...
-	shareName   = flag.String("shareName", "", "Name of the share")
-	targetPath  = flag.String("targetPath", "", "Path to mount volume at")
-	smbUsername = flag.String("smbUsername", "", "WINDOWS ONLY: Username to login to SMB share")
-	smbPassword = flag.String("smbPassword", "", "WINDOWS ONLY: Password to login to SMB share")
+Supported flags:
+`
 
-	request = flag.String("request", "", requestHelpMessage)
+func main() {
+	flag.CommandLine = flag.NewFlagSet("test", flag.ExitOnError)
```
msau42: is this required? Is "-h" sufficient?

hack/csi_client/cmd/main.go
```diff
@@ -1,34 +1,45 @@
...
+	flag.ErrHelp = errors.New("flag: help req")
+	flag.CommandLine.Usage = func() {
```
msau42: Is putting everything under `CommandLine` required? From the documentation, it seems like the variables are top-level in the pkg

hack/csi_client/cmd/main.go
almos98: Sorry I forgot to remove this.

hack/csi_client/cmd/main.go
almos98: You're right, it doesn't. Didn't catch this earlier.

msau42:
/approve

Can you also add a README into hack/csi-client, and squash the commits?

almos98:
Done. Could you take a look at the README and see if it's good? @pjh @msau42

PR #39, opened 2019-09-04
Make CSI Client more generic

almos98:
By replacing the `flag` library with [`pflag`](https://github.com/spf13/pflag), Volume Attributes and Secrets can be passed in instead of using specific flags: `shareAddr`, `shareName`, `smbUsername` and `smbPassword`.

The fields required in these maps are the exact same as required by the driver. Eg: in [gcp-filestore-csi-driver](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/19149de5fdf104b0f4ae221792d43fbffd6a1283/pkg/csi_driver/node.go#L177), the volume attributes should have the fields "ip" and "volume".

Example usage:
```Powershell
.\csi-client.exe --endpoint "/tmp/csi.sock" --request "NodePublish" --volumeAttr "ip=localhost,volume=SMBShare" --targetPath "C:\mountedShare" --secrets "smbUser=foo\bar,smbPassword=foobar"
```

There's two things I'm still working on:
1. The usage isn't really working as it was before replacing `flag` library.
2. Commas (`,`) can't be passed in map keys or values and will cause an error.


pjh:
cc @pjh @adjackura

almos98:
Fixed the usage message. Could you take a look @msau42  @pjh

hack/csi_client/pkg/csi/client.go
```diff
@@ -24,9 +24,11 @@ const (
 
 var (
 	goOs          = runtime.GOOS
+	errVolumeAttr = fmt.Errorf("volume attributes are required")
```
msau42: If we want this to be generic, then I think volumeattr and secrets should be optional

hack/csi_client/pkg/csi/client.go
```diff
@@ -24,9 +24,11 @@ const (
...
 	errHost       = fmt.Errorf("host name is required")
```
msau42: Are all of these errs still needed?

hack/csi_client/cmd/main.go
```diff
@@ -14,7 +15,7 @@ Usage of %s:
 	Other necessary flags are determined based on the request type.
 
 	NodePublish
-		- shareAddr, shareName, targetPath, smbUsername (WINDOWS), smbPassword (WINDOWS)
+		- volumeAttr, targetPath, secrets (Windows)
```
msau42: Can you update the README with an example for Filestore?

hack/csi_client/pkg/csi/client.go
almos98: No they are not, I forgot to remove them.

hack/csi_client/pkg/csi/client.go
almos98: Then should I remove `validateRequest`? That won't be doing much.

hack/csi_client/cmd/main.go
almos98: Done.

almos98:
I removed `client_test.go` as the only function it was testing has been removed (`validateRequest`).

Ready for review again @msau42 @pjh @adjackura

hack/csi_client/pkg/csi/client.go
```diff
@@ -24,11 +24,9 @@ const (
```
pjh: Looks like some of the constants above are now unused - delete them?

hack/csi_client/pkg/csi/client.go
pjh: Looks like some of these vars are now unused - delete them?

hack/csi_client/pkg/csi/client.go
```diff
@@ -114,17 +74,11 @@ func (c *Client) NewRequest(req *Request) (err error) {
...
-			},
+			VolumeAttributes: req.VolumeAttr,
 		}
 
 		if goOs == "windows" {
-			csiReq.NodePublishSecrets = map[string]string{
-				smbUser:     fmt.Sprintf("%s\\%s", req.ShareAddr, req.Username),
-				smbPassword: req.Password,
-			}
+			csiReq.NodePublishSecrets = req.Secrets
```
pjh: Does this have to be only for Windows? Could this be useful (or at least not-harmful) for requests on Linux too?

hack/csi_client/pkg/csi/client.go
```diff
@@ -114,17 +74,11 @@ func (c *Client) NewRequest(req *Request) (err error) {
 					Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 				},
 			},
-			VolumeAttributes: map[string]string{
-				attrIP:     req.ShareAddr,
-				attrVolume: req.ShareName,
-			},
+			VolumeAttributes: req.VolumeAttr,
```
pjh: I like that this PR makes the code both more useful and simpler :)

hack/csi_client/pkg/csi/client.go
almos98: Secrets are ignored by anything that doesn't need them, so I guess not harmful.

hack/csi_client/pkg/csi/client.go
```diff
@@ -14,21 +14,15 @@ import (
...
-	smbPassword      = "smbPassword"
 	reqNodePublish   = "nodepublish"
 	reqNodeUnpublish = "nodeunpublish"
 )
 
 var (
 	goOs          = runtime.GOOS
-	errHost       = fmt.Errorf("host name is required")
-	errVolume     = fmt.Errorf("volume name is required")
+	errVolumeAttr = fmt.Errorf("volume attributes are required")
```
msau42: volumeAttr and secret errs are not used either.  Should we error for empty targetPath?

hack/csi_client/README.md
```diff
@@ -16,9 +16,7 @@ As of now, only two types of requests are supported:
 
 If the SMB share requires a username and password, these can be provided by using the flags `smbUsername` and `smbPassword`:
 ```Powershell
-.\csi-client.exe --endpoint="C:\tmp\csi.sock" --request="nodepublish" --shareAddr="localhost"
---shareName="SMBShare" --targetPath="C:\SMBShareMount"
---smbUsername="smbuser" --smbPassword="foobar"
+.\csi-client.exe --endpoint="C:\tmp\csi.sock" --request="nodepublish" -- volumeAttr="ip=localhost,volume=SMBShare" --targetPath="C:\SMBShareMount" --secrets="smbUser=smbuser,smbPassword=foobar"
```
msau42: Can you add a note in the README about the limitation that comma is used as a delimeter and therefore cannot be a character in the secret?

hack/csi_client/pkg/csi/client.go
almos98: True, I'll just remove all errors. Since we're leaving all validation to the driver, I don't think erroring on an empty targetPath is necessary. The driver will return an error on empty targetPath.

hack/csi_client/pkg/csi/client.go
```diff
@@ -14,21 +14,12 @@ import (
...
 )
 
 var (
-	goOs          = runtime.GOOS
-	errHost       = fmt.Errorf("host name is required")
-	errVolume     = fmt.Errorf("volume name is required")
-	errTargetPath = fmt.Errorf("targetPath is required")
-	errNoUsername = fmt.Errorf("username is required")
-	errNoPassword = fmt.Errorf("password is required")
+	goOs = runtime.GOOS
```
msau42: do we use this anywhere?

almos98: It was stubbed out for testing but I don't think there's OS-specific code anymore. I'll review it.

msau42:
/lgtm
Thanks!

PR #40, opened 2019-09-25
Enable Cloud Filestore API when setting up for a new project.

axisofentropy:
Also update the path for the example `nginx` Pod.


msau42:
/ok-to-test
/lgtm
/approve
Thanks!

PR #51, opened 2020-07-20
Update Makefile with image tags

saikat-royc:
This patch makes changes to the makefile
1. Dynamically generate image tags
2. Update the image build and push rules for all tags.
3. Create a cloudbuild yaml to enable auto builds.

TODO:
Add a prow job based on the cloudbuild.yaml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Build driver
+# Build driver go binary
 FROM golang:1.14.4 as builder
 WORKDIR /go/src/sigs.k8s.io/gcp-filestore-csi-driver
 ADD . .
-ARG TAG
-RUN CGO_ENABLED=0 go build -a -ldflags '-X main.version='"${TAG:-latest}"' -extldflags "-static"' -o bin/gcp-filestore-csi-driver ./cmd
+RUN make driver
 
 # Install nfs packages
 FROM launcher.gcr.io/google/debian9 as deps
@@ -83,7 +82,9 @@ RUN apt-get autoremove -y && \
 
 # Copy driver into image
 FROM deps
-COPY --from=builder /go/src/sigs.k8s.io/gcp-filestore-csi-driver/bin/gcp-filestore-csi-driver /gcp-filestore-csi-driver
+ARG DRIVERBINARY
+COPY --from=builder /go/src/sigs.k8s.io/gcp-filestore-csi-driver/bin/${DRIVERBINARY} /${DRIVERBINARY}
 COPY deploy/kubernetes/nfs_services_start.sh /nfs_services_start.sh
+
 
 ENTRYPOINT ["/gcp-filestore-csi-driver"]
--- a/Makefile
+++ b/Makefile
@@ -12,30 +12,71 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMAGE=gcr.io/$(PROJECT)/gcp-filestore-csi-driver
+# Core Filestore CSI driver binary
+DRIVERBINARY=gcp-filestore-csi-driver
 
+# A space-separated list of image tags under which the current build is to be pushed.
+# Determined dynamically.
+STAGINGVERSION=
+ifdef GCP_FS_CSI_STAGING_VERSION
+	STAGINGVERSION=${GCP_FS_CSI_STAGING_VERSION}
+else
+	STAGINGVERSION=$(shell ./build/generate_image_tags.sh)
+endif
+$(info STAGINGVERSION is $(STAGINGVERSION))
+
+STAGINGIMAGE=
+ifdef GCP_FS_CSI_STAGING_IMAGE
+	STAGINGIMAGE=$(GCP_FS_CSI_STAGING_IMAGE)
+else
+	STAGINGIMAGE=gcr.io/$(PROJECT)/gcp-filestore-csi-driver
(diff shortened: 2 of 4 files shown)


mattcary:
Out of curiosity, where was all this copied from? I assume it's boilerplate.

Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM golang:1.14.4 as builder
 WORKDIR /go/src/sigs.k8s.io/gcp-filestore-csi-driver
 ADD . .
 ARG TAG
-RUN CGO_ENABLED=0 go build -a -ldflags '-X main.version='"${TAG:-latest}"' -extldflags "-static"' -o bin/gcp-filestore-csi-driver ./cmd
+RUN CGO_ENABLED=0 go build -a -mod=vendor -ldflags '-X main.version='"${TAG:-latest}"' -extldflags "-static"' -o bin/gcp-filestore-csi-driver ./cmd
```
msau42, line 20: You could run make from the Makefile so the build commands aren't duplicated

saikat-royc:
@mattcary   I looked at how PD CSI driver generates tags. (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/Makefile). This however does not generate canary tags.

For logic for generating canary tags looked at the external-snapshotter, looked at how IMAGE_TAGS are generated. (snapshotter uses a full fleged release tools to generate images). However for filestore checked with @msau42  and I have put very basic logic in the Makefile to generate and push "canary" and "vx.x.x", or "commit hash" images in k8s-staging-cloud-provider-gcp. This is so that I can update the deploy/k8s/manifests for the FS driver to use the canary images.

Further to understand how cloudbuild.yaml works read the documentation at https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md.
And the image builder code (used in the cloudbuild.yaml) https://github.com/kubernetes/test-infra/tree/master/images/builder on how the builder parses the yaml

Dockerfile
saikat-royc: Ack

mattcary:
Thanks! This is a useful set of breadcrumbs.

> @mattcary I looked at how PD CSI driver generates tags. (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/Makefile). This however does not generate canary tags.
> 
> For logic for generating canary tags looked at the external-snapshotter, looked at how IMAGE_TAGS are generated. (snapshotter uses a full fleged release tools to generate images). However for filestore checked with @msau42 and I have put very basic logic in the Makefile to generate and push "canary" and "vx.x.x", or "commit hash" images in k8s-staging-cloud-provider-gcp. This is so that I can update the deploy/k8s/manifests for the FS driver to use the canary images.
> 
> Further to understand how cloudbuild.yaml works read the documentation at https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md.
> And the image builder code (used in the cloudbuild.yaml) https://github.com/kubernetes/test-infra/tree/master/images/builder on how the builder parses the yaml
> 
> Checked that make push works:
> https://pantheon.corp.google.com/gcr/images/k8s-staging-cloud-provider-gcp/GLOBAL/gcp-filestore-csi-driver

Dockerfile
```diff
@@ -83,7 +80,10 @@ RUN apt-get autoremove -y && \
 
 # Copy driver into image
 FROM deps
-COPY --from=builder /go/src/sigs.k8s.io/gcp-filestore-csi-driver/bin/gcp-filestore-csi-driver /gcp-filestore-csi-driver
+ARG DRIVERBINARY
+ARG binary=./bin/${DRIVERBINARY}
+COPY ${binary} /${DRIVERBINARY}
```
mattcary, line 85: Is there any disadvantage to not running the makefile in the docker build and instead running externally and then copying in as you do?

saikat-royc: I thought about calling make from the dockerfile. However, for an automatic cloudbuild the entry point is make  build-image-and-push (https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/51/files#diff-e5fc65747fb9417968a3295dc1ac5111R25).
i.e make would trigger docker build. Calling make again from the dockerfile seem to be cyclic. Hence I created separate rules in make file, and which would be called in the following order:
1. driver (which builds the go binary, https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/51/files#diff-e5fc65747fb9417968a3295dc1ac5111R25)
2. image (which then builds the docker image, https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/51/files#diff-b67911656ef5d18c4ae36cb6741b7965R45)
3. build-image-and-push (which pushes the docker image to a registry, https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/51/files#diff-b67911656ef5d18c4ae36cb6741b7965R73)

msau42: The advantage of building in the docker build is then it removes the requirement of installing the right version of golang locally. But I guess that doesn't really help if you also want to do things like run unit tests during development.

mattcary: Not having to configure go binaries is a big win IMHO, even if extra work is needed for testing it means that just deploying a driver from master is easy.

The image build would call `make driver`, and the image rule wouldn't have to depend on the driver rule.

saikat-royc: Done

mattcary:
/lgtm

thx

PR #55, opened 2020-07-24
Add resizer feature to Filestore driver

### LINKED ISSUE #54: Support Volume Expansion for Filestore CSI driver
Support Volume Expansion for Filestore CSI driver

saikat-royc:
1. add rbac rules and resizer sidecar
2. Add code for controller expand volume
3. updated exisitng unit tests

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Add support for Volume Expansion 

Fixes #54 

```release-note
Add support for Volume Expansion for Filestore CSI driver
```

### DIFF
--- a/examples/kubernetes/sc-latebind.yaml
+++ b/examples/kubernetes/sc-latebind.yaml
@@ -6,3 +6,4 @@ provisioner: filestore.csi.storage.gke.io
 parameters:
   location: us-central1-c
 volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -57,6 +57,7 @@ type Service interface {
 	DeleteInstance(ctx context.Context, obj *ServiceInstance) error
 	GetInstance(ctx context.Context, obj *ServiceInstance) (*ServiceInstance, error)
 	ListInstances(ctx context.Context, obj *ServiceInstance) ([]*ServiceInstance, error)
+	ResizeInstance(ctx context.Context, obj *ServiceInstance) (*ServiceInstance, error)
 }
 
 type gcfsServiceManager struct {
@@ -69,6 +70,9 @@ const (
 	locationURIFmt  = "projects/%s/locations/%s"
 	instanceURIFmt  = locationURIFmt + "/instances/%s"
 	operationURIFmt = locationURIFmt + "/operations/%s"
+
+	// Patch update masks
+	fileShareUpdateMask = "file_shares"
 )
 
 var _ Service = &gcfsServiceManager{}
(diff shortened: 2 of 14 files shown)


examples/kubernetes/sc-latebind.yaml
```diff
@@ -6,3 +6,4 @@ provisioner: filestore.csi.storage.gke.io
 parameters:
   location: us-central1-c
 volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
```
mattcary, line 9: And add to demo-sc.yaml?

pkg/cloud_provider/file/file.go
```diff
@@ -251,6 +255,64 @@ func (manager *gcfsServiceManager) ListInstances(ctx context.Context, obj *Servi
...
+	if err != nil {
+		glog.Errorf("Failed to get instance %s for resize operation, error %v", instanceuri, err)
+		return nil, err
+	}
+
+	// High Scale tier supports shrink of capacity. However CSI spec does not support it.
+	if obj.Volume.SizeBytes <= instance.Volume.SizeBytes {
+		return instance, nil
+	}
+
```
mattcary, line 270: So regular tier suports changes in 1G increments, and high scale in 10T increments:
https://cloud.google.com/filestore/docs/editing-instances

I don't know if the API errors or rounds if you try to change in a smaller increment. Any ideas on how to handle that?

saikat-royc: Thats a good question. The current spec for Fileshare only provides CapacityGb option (https://cloud.google.com/filestore/docs/reference/rest/v1beta1/projects.locations.instances#fileshareconfig). i.e we have option to either do a floor or ceil of the requested bytes. Current implementation does a ceil.
I verified that, say we increase a PVC size to from 1Tb to 1Tb and another 500 mb, the driver sends a request for 1Tb + 1Gb, and the filestore edit API succeeds. This means now PVC reports 1101109657600 bytes (~1025.5 gb) while PV reports 1026 Gb.

If we go with floor option, that seems more consistent. the 1Tb + 500mb request would be floor to 1Tb and no resize would be done. I see PD does a floor too.(https://sourcegraph.com/github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver@82298d110dfe8acddce893825b553a62c7fb37b9/-/blob/pkg/gce-cloud-provider/compute/gce-compute.go#L663)

Regarding high scale tier, I dont have a setup to test. I expect any edits <10Tb should return an error. So no extra logic should be needed in the driver, and let filestore give back the appropriate error. But will confirm with the high scale team.

examples/kubernetes/sc-latebind.yaml
saikat-royc: right will do.

pkg/cloud_provider/file/file.go
mattcary: Ok, makes sense. Generally reporting back filestore error makes sense, we shouldn't encode their logic in our driver.

PR #56, opened 2020-08-04
Topology feature support for Filestore CSI driver

saikat-royc:
1. Support Topology constraints feature in FS CSI driver
2. Use Storage Class allowedTopology field to provide input to
   FS driver for instance creation.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #

This patch removes the need to 'location' parameter in storage class, and uses CreateVolumeRequest.accessibility_requirements  field to determine a zone for instance creation. Further the CreateVolume response does not put any restriction on the volume, since it can be accessed across all zones on the same VPC.

```release-note
Topology feature support for Filestore CSI driver
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,26 @@ volume. Customizable parameters for volume creation include:
 | ---------------   | ----------------------- |-----------                             | ----------- |
 | tier              | "standard"<br>"premium" | "standard"                             | storage performance tier |
 | network           | string                  | "default"                              | VPC name |
-| location          | string                  | zone where the plugin<br>is running in | zone |
 | reserved-ipv4-cidr| string		              | ""                                     | CIDR range to allocate Filestore IP Ranges from.<br>The CIDR must be large enough to accommodate multiple Filestore IP Ranges of /29 each |
 
 For Kubernetes clusters, these parameters are specified in the StorageClass.
 
 Note that non-default networks require extra [firewall setup](https://cloud.google.com/filestore/docs/configuring-firewall)
+
+## Current supported Features
+* Volume resizing: CSI Filestore driver supports volume expansion for all supported Filestore tiers.
+* Topology preferences: Filestore performance and network usage is affected by topology. For example, it is recommended to run
+  workloads in the same zone where the Cloud Filestore instance is provisioned in. The following table describes how provisioning can be tuned by topology. The volumeBindingMode is specified in the StorageClass used for provisioning. 'strict-topology' is a flag passed to the CSI provisioner sidecar. 'allowedTopology' is also specified in the StorageClass. The Filestore driver will use the first topology in the preferred list, or if empty the first in the requisite list. If topology feature is not enabled in CSI provisioner (--feature-gates=Topology=false), CreateVolume.accessibility_requirements will be nil, and the driver simply creates the instance in the zone where the driver deployment running.
+
+
+  | SC Bind Mode         | 'strict-topology' | SC allowedTopology  | CSI provisioner Behavior    |
+  | -------------------- | ----------------- | ------------------- | --------------------------- |
+  | WaitForFirstCustomer |       true        |        Present      | If the topology of the node selected by the schedule is not in allowedTopology, provisioning fails and the scheduler will continue with a different node. Otherwise, CreateVolume is called with requisite and preferred topologies set to that of the selected node |
+  | WaitForFirstCustomer |       false       |        Present      | If the topology of the node selected by the schedule is not in allowedTopology, provisioning fails and the scheduler will continue with a different node. Otherwise, CreateVolume is called with requisite set to allowedTopology and preferred set to allowedTopology rearranged with the selected node topology as the first parameter |
+  | WaitForFirstCustomer |       true        |        Not Present  | Call CreateVolume with requisite set to selected node topology, and preferred set to the same |
+  | WaitForFirstCustomer |       false       |        Not Present  | Call CreateVolume with requisite set to aggregated topology across all nodes, which matches the topology of the selected node, and preferred is set to the sorted and shifted version of requisite, with selected node topology as the first parameter |
+  | Immediate            |       N/A         |        Present      | Call CreateVolume with requisite set to allowedTopology and preferred set to the sorted and shifted version of requisite at a randomized index |
+  | Immediate            |       N/A         |        Not Present  | Call CreateVolume with requisite = aggregated topology across nodes which contain the topology keys of CSINode objects, preferred = sort and shift requisite at a randomized index |
 
 ## Future Features
 * Non-root access: By default, GCFS instances are only writable by the root user
@@ -58,11 +72,6 @@ Note that non-default networks require extra [firewall setup](https://cloud.goog
   [nfs-client](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs-client)
   external provisioner can be used to provide similar functionality for
   Kubernetes clusters.
-* Volume resizing: CSI Filestore driver supports volume expansion for all supported Filestore tiers.
-* Topology preferences: For better performance, it is recommended to run
-  workloads in the same zone where the Cloud Filestore instance is provisioned in. In the
-  future, the location where to create a Cloud Filestore instance could be automatically
-  influenced by where the workload is scheduled.
 
 ## Kubernetes User Guide
 1. One-time per project: Create GCP service account for the CSI driver and set the Cloud
--- a/examples/kubernetes/topology/immediate-binding/demo-deployment-immediate-allowedtopo.yaml
+++ b/examples/kubernetes/topology/immediate-binding/demo-deployment-immediate-allowedtopo.yaml
@@ -1,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: web-server-immediate-allowedtopo
+  labels:
+    app: nginx
+spec:
+  replicas: 5
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx
+        volumeMounts:
+        - mountPath: /usr/share/nginx/html
+          name: mypvc
+      nodeSelector:
+        topology.gke.io/zone: us-central1-a
+      volumes:
+      - name: mypvc
+        persistentVolumeClaim:
+          claimName: test-pvc-fs-immediate-binding-allowedtopo
+
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: test-pvc-fs-immediate-binding-allowedtopo
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: csi-filestore-immediate-binding-allowedtopo
+  resources:
+    requests:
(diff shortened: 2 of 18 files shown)


examples/kubernetes/sc-allowedtopo.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-filestore-allowedtopo
+provisioner: filestore.csi.storage.gke.io
+volumeBindingMode: WaitForFirstConsumer
+allowVolumeExpansion: true
+allowedTopologies:
```
msau42, line 8: Do you see a need for a use case where someone wants to provision an instance in a different zone as the pod?

Currently the scheduling logic will cause the Pod to be scheduled in the [same zones](https://github.com/kubernetes/kubernetes/blob/3647766cbcb39558291568b5cd0e38ad34a9feeb/pkg/controller/volume/scheduling/scheduler_binder.go#L865) during the first provisioning.

saikat-royc: I am not sure of such a case. However, allowedTopologies will allow user to restrict the candidate zones where a FS instance can be deployed, which this example highlights. say for an regional cluster, user can provide a subset of zones where the FS instance can be deployed.(although it will be still accesible to the entities in the same VPC)

README.md
```diff
@@ -60,9 +59,10 @@ Note that non-default networks require extra [firewall setup](https://cloud.goog
   Kubernetes clusters.
 * Volume resizing: CSI Filestore driver supports volume expansion for all supported Filestore tiers.
 * Topology preferences: For better performance, it is recommended to run
-  workloads in the same zone where the Cloud Filestore instance is provisioned in. In the
-  future, the location where to create a Cloud Filestore instance could be automatically
-  influenced by where the workload is scheduled.
+  workloads in the same zone where the Cloud Filestore instance is provisioned in. Zone preference
```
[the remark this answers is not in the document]
mattcary: +1 to moving this out of future features.

I'd also be a bit more explicit about how it works, with particular examples, so that it's clear what the order of preferences is. eg, (IIUC), specifying allowedTopologies=A,B and then binding with WaitForFirstConsumer=C will result in an instance in C; if binding is immediate then the topology would be A, etc.

examples/kubernetes/sc-allowedtopo.yaml
mattcary: Wouldn't the flow in this case be to create a PV via a bindingMode=ImmediatePVC and a requisite topology specifying zone X for the filestore instance, then a pod with an affinity to a node pool in zone Y?

We could document that in the examples or docs (assuming I got it right...)

I have a hard time seeing how one could make this any simpler in practice without extending the existing topology representation (?)

msau42: Yes, I think immediate binding and setting allowed topologies is the way to address this scenario.

README.md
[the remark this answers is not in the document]
saikat-royc: I have created a matrix of filestore create volume behavior based on the various knobs available. please take a look

README.md
```diff
@@ -41,13 +41,39 @@ volume. Customizable parameters for volume creation include:
...
+  For creating the instance, Filestore driver picks the first topology in the preferred list, or the
+  first topology of the requisite list (if preferred list is empty). If topology feature is not enabled in
+  CSI provisioner (--feature-gates=Topology=false),  CreateVolume.accessibility_requirements will be nil, and the driver
+  simply creates the instance in the zone where its running.
+
+
+| SC Bind Mode         | 'strict-topology' | SC allowedTopology  | CSI provisioner Behavior    |
+| -------------------- | ----------------- | ------------------- | --------------------------- |
+| WaitForFirstCustomer |       true        |        Present      | CSI provisioner fails the provisioning if the topology of the node selected by the scheduler does not fall under SC allowedTopology list (scheduler tries to select a different node). Else, call CreateVolume with requisite = the topology of the selected node, Preferred = Requisite |
+| WaitForFirstCustomer |       false       |        Present      | Call CreateVolume with requisite = SC allowedTopology, preferred = sort and shift requiste list with selected node topology as the first parameter |
```
mattcary, line 71: nit: isn't it "requisite"? (two i's)

saikat-royc: yes will fix it

README.md
```diff
@@ -41,13 +41,39 @@ volume. Customizable parameters for volume creation include:
...
+  CSI provisioner side car 'strict-topology' flag, SC allowedTopology field. The 'CSI provisioner Behavior' column describes
+  how the CreateVolume.accessibility_requirements.requisite and CreateVolume.accessibility_requirements.preferred is populated
+  by the CSI provisioner.
+  For creating the instance, Filestore driver picks the first topology in the preferred list, or the
+  first topology of the requisite list (if preferred list is empty). If topology feature is not enabled in
+  CSI provisioner (--feature-gates=Topology=false),  CreateVolume.accessibility_requirements will be nil, and the driver
+  simply creates the instance in the zone where its running.
+
+
+| SC Bind Mode         | 'strict-topology' | SC allowedTopology  | CSI provisioner Behavior    |
```
mattcary, line 68: Clean up the headings to match the previous paragraph

README.md
```diff
@@ -41,13 +41,39 @@ volume. Customizable parameters for volume creation include:
...
+| -------------------- | ----------------- | ------------------- | --------------------------- |
+| WaitForFirstCustomer |       true        |        Present      | CSI provisioner fails the provisioning if the topology of the node selected by the scheduler does not fall under SC allowedTopology list (scheduler tries to select a different node). Else, call CreateVolume with requisite = the topology of the selected node, Preferred = Requisite |
```
mattcary, line 70: nit: maybe more clear to write "If the topology if the node selected by the schedule is not in `allowedTopology`, provisioning fails and the scheduler will continue with a different node. Otherwise, CreateVolume is called with requisite and preferred topologies set to that of the selected node."

README.md
mattcary, line 71: Matching the above, "Call CreateVolume with requisite set to `allowedTopology` and preferred set to `allowedTopology` rearranged with the selected node topology as the first parameter. The node topology will be added if it is not in `allowedTopology`.

README.md
```diff
@@ -41,13 +41,39 @@ volume. Customizable parameters for volume creation include:
...
+| WaitForFirstCustomer |       false       |        Present      | Call CreateVolume with requisite = SC allowedTopology, preferred = sort and shift requiste list with selected node topology as the first parameter |
+| WaitForFirstCustomer |       true        |        Not Present  | Call CreateVolume with requisite =  Selected node topology, preferred = requiste |
+| WaitForFirstCustomer |       false       |        Not Present  | Call CreateVolume with requisite = aggregated topology across all nodes, which match the topology of the selected node, preferred = sort and shift Requiste with selected node topology as the first parameter |
```
mattcary, line 73: reword the rest of these to match the above.

README.md
saikat-royc: from [code](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/topology.go#L180) I see that "The node topology will NOT be added if it is not in allowedTopology".  As seen in code, selectedNodeTopology is not used, for waitForFirstCustomer, strict-topology  = false and allowedTopologies present.  requisite list is populated directly from allowedTopogies.

only the following statement is true:
"Call CreateVolume with requisite set to allowedTopology and preferred set to allowedTopology rearranged with the selected node topology as the first parameter"

README.md
saikat-royc: Done

README.md
saikat-royc: Done

README.md
```diff
@@ -41,13 +41,39 @@ volume. Customizable parameters for volume creation include:
...
-| location          | string                  | zone where the plugin<br>is running in | zone |
 | reserved-ipv4-cidr| string		              | ""                                     | CIDR range to allocate Filestore IP Ranges from.<br>The CIDR must be large enough to accommodate multiple Filestore IP Ranges of /29 each |
 
 For Kubernetes clusters, these parameters are specified in the StorageClass.
 
 Note that non-default networks require extra [firewall setup](https://cloud.google.com/filestore/docs/configuring-firewall)
 
+## Current supported Features
+* Volume resizing: CSI Filestore driver supports volume expansion for all supported Filestore tiers.
+* Topology preferences: For better performance, it is recommended to run
```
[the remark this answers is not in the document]
saikat-royc: Done

README.md
saikat-royc: Done

examples/kubernetes/sc-allowedtopo.yaml
saikat-royc: Added 1 example each for both delayed binding and immediate binding. The other scenarios are listed in the behavior matrix in README

examples/kubernetes/topology/immediate-binding/demo-deployment-immediate-allowedtopo.yaml
```diff
@@ -0,0 +1,29 @@
...
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx
+        volumeMounts:
+        - mountPath: /usr/share/nginx/html
+          name: mypvc
+      nodeSelector:
```
saikat-royc, line 23: node affinity set here based on allowedTopogy in SC

README.md
mattcary: Ok, thanks for confirming. I think this supports the decision to go into some detail on how things work, clearly it's complicated.

examples/kubernetes/topology/immediate-binding/pvc-immediate.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
```
mattcary, line 1: Why not put this in the same yaml as the immediate deployment, as you've done with the delayed deployment? I find that it's easier to navigate fewer files, if possible.

saikat-royc: Done

README.md
```diff
@@ -41,13 +41,27 @@ volume. Customizable parameters for volume creation include:
...
+  | WaitForFirstCustomer |       false       |        Present      | If the topology of the node selected by the schedule is not in allowedTopology, provisioning fails and the scheduler will continue with a different node. Otherwise, CreateVolume is called with requisite set to allowedTopology and preferred set to allowedTopology rearranged with the selected node topology as the first parameter |
+  | WaitForFirstCustomer |       true        |        Not Present  | Call CreateVolume with requisite set to selected node topology, and preferred set to the same |
+  | WaitForFirstCustomer |       false       |        Not Present  | Call CreateVolume with requisite set to aggregated topology across all nodes, which match the topology of the selected node, and preferred is set to the sorted and shifted version of requisite, with selected node topology as the first parameter |
```
mattcary, line 61: nit: s/match/matches in "which *matches* the topology of the selected node"

saikat-royc: Done

PR #64, opened 2020-10-08
Update Libraries and use new File service creation API

xvzf:
**What type of PR is this?**
/kind cleanup
> /kind feature

This update allows using WorkloadIdentity instead of passing an access token to the controller preventing secret leakage.

```release-note
- Added support for WorkloadIdentity
```


mattcary:
Can you change the name of this PR to "Update Libraries and use new File service creation API"?

OK to leave the workload identity motivation in the description as-is.

@saikat-royc AFAIU this will work for our node scope ADC access for k8s-on-gce testing, can you confirm that?

saikat-royc:
@mattcary @xvzf  I am not sure if removing the newOauthClient() would work.
newOauthClient() uses [FindDetaultCredentials](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/88725951352362b856ec760b883d1b8ce227b77e/vendor/golang.org/x/oauth2/google/default.go#L76), which first checks for GOOGLE_APPLICATION_CREDENTIALS env variable and reads credentials from the file specified in the env variable. The variable is setup here in the [manifest](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/deploy/kubernetes/manifests/controller.yaml#L54)

For using ADC(Application default credentials) on GCE, the GOOGLE_APPLICATION_CREDENTIALS env will not be provided in the manifest, so  and return default credentials
```
DefaultCredentials{
			ProjectID:   id,
			TokenSource: ComputeTokenSource("", scopes...),
		}, nil
```
[here](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/88725951352362b856ec760b883d1b8ce227b77e/vendor/golang.org/x/oauth2/google/default.go#L107).

mattcary:
The documentation for NewService suggests that it does use ADC but I haven't tried it myself, see vendor/google.golang.org/api/file/v1/file-gen.go

saikat-royc:
> The documentation for NewService suggests that it does use ADC but I haven't tried it myself, see vendor/google.golang.org/api/file/v1/file-gen.go

ADC may work. I will try to apply this patch locally and check that we can provision filestore using ADC. But in this patch , we are loosing ability to pick up credentials from GOOGLE_APPLICATION_CREDENTIALS. atleast from a first look, it looks like that.

msau42:
Double check what pd csi driver does.

saikat-royc:
> Double check what pd csi driver does.

pdcsi driver also creates token source and creates newOauthClient with that tokensource
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce.go#L85
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce.go#L180

mattcary:
The API seems to be a bit different; in the pd csi driver we create a token source and pass in; if the env variable in generateTokenSource is empty then the default token source does the right thing and picks up application default creds from the metadata server.

The filestore API used in this PR allows you to add a tokenSource to the NewService() call, but if you don't do that (ie just call NewService(ctx)) the documentation says that the application default creds will be used which should be the right thing.

If we need to add the same altTokenSource as in the pd csi driver we should be able to add the appropriate token to the NewService() call I think.

xvzf:
I have never used it with the first API but when reviewing the [docs](https://cloud.google.com/docs/authentication/production#auth-cloud-explicit-go) we're going first with `GOOGLE_APPLICATION_CREDENTIALS` and in case it is not set it goes to the attached SA.

It works on other APIs. I'll try to find some time to verify the behaviour.

mattcary:
Sorry for the delay in getting back to you.

The overlay change Saikat is working on (#65) looks like it will work fine with what you're doing here so let's merge it.

/lgtm

xvzf:
Thanks @mattcary; will fix the CI tomorrow!

First time contributing - will it auto merge?

mattcary:
Basically. It looks like you have to push some changes to fix the unit
tests; that will need another lgtm from me, but once it happens it will
automerge without your needing to hit a button or whatever.

xvzf:
@mattcary feel free to check; should work!

xvzf:
Any update here?

mattcary:
/lgtm

Sorry, I was out of office and this thread got buried!

PR #66, opened 2020-10-14
Fix CSI e2e test

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
/kind failing-test
> /kind feature
> /kind flake

Filestore driver has added new capabilities. The patch ensures the vanilla CSI single zone e2e test passes with the new changes.

Fixes #

Tested manually by deploying a GCE cluster
1. PROJECT=saikatroyc-gke-dev DEPLOY_VERSION=dev ./deploy/project_setup.sh
2. NODE_SCOPES=https://www.googleapis.com/auth/cloud-platform NODE_SERVICE_ACCOUNT=gcp-filestore-csi-driver-sa@saikatroyc-gke-dev.iam.gserviceaccount.com kubetest --up
3. PROJECT=saikatroyc-gke-dev GCFS_IAM_NAME=gcp-filestore-csi-driver-sa@saikatroyc-gke-dev.iam.gserviceaccount.com ./test/run_e2e_local.sh

```release-note
Fix CSI e2e test
```

### DIFF
--- a/test/e2e/tests/single_zone_e2e_test.go
+++ b/test/e2e/tests/single_zone_e2e_test.go
@@ -29,13 +29,11 @@ import (
 
 const (
 	testNamePrefix = "gcfs-csi-e2e-"
-
-	defaultSizeGb int64 = 5
 )
 
 var _ = Describe("Google Cloud Filestore CSI Driver", func() {
 
-	It("Should create->mount volume and check if it is writable, then unmount->delete and check if volume is deleted", func() {
+	It("Should create->stage->mount volume and check if it is writable, then unmount->unstage->delete and check if volume is deleted", func() {
 		Expect(testInstances).NotTo(BeEmpty())
 		testContext, err := testutils.GCFSClientAndDriverSetup(testInstances[0])
 		Expect(err).To(BeNil(), "Set up new Driver and Client failed with error")
@@ -59,9 +57,14 @@ var _ = Describe("Google Cloud Filestore CSI Driver", func() {
 		}()
 
 		// TODO validate the Filestore instance creation at the cloud provider layer
+		stageDir := filepath.Join("/tmp/", volName, "stage")
+		_, err = instance.SSH(fmt.Sprint("mkdir -p ", stageDir))
+
+		err = client.NodeStageVolume(vol.GetVolumeId(), stageDir, vol.GetVolumeContext())
+		Expect(err).To(BeNil(), "NodeStageVolume failed with error")
+
 		// Mount Disk
 		publishDir := filepath.Join("/tmp/", volName, "mount")
-
 		// Make remote directory
 		_, err = instance.SSH(fmt.Sprint("mkdir -p ", publishDir))
 		Expect(err).To(BeNil(), "Failed to delete remote directory")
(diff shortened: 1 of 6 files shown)


saikat-royc:
I manually triggered the e2e test for prow to check the behavior. The test failed because the filestore API is not enabled for the project
```
CreateVolume failed with error: rpc error: code = Internal desc = googleapi: Error 403: Cloud Filestore API has not been used in project 649641406145 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/file.googleapis.com/overview?project=649641406145 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
```
@msau42  thoughts?

mattcary:
> I manually triggered the e2e test for prow to check the behavior. The test failed because the filestore API is not enabled for the project
> 
> ```
> CreateVolume failed with error: rpc error: code = Internal desc = googleapi: Error 403: Cloud Filestore API has not been used in project 649641406145 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/file.googleapis.com/overview?project=649641406145 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
> ```
> 
> It passes locally for me, because api was enabled by [project_setup.sh](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/deploy/project_setup.sh) script.
> @msau42 thoughts?

Maybe contact engprod, (google-internal go/kubernetes-engprod).

msau42:
we need to add the filestore api to the project fixer/reconciler for the boskos pools that we're using. agree engprod can help us here.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -59,9 +57,14 @@ var _ = Describe("Google Cloud Filestore CSI Driver", func() {
 		}()
 
 		// TODO validate the Filestore instance creation at the cloud provider layer
+		stageDir := filepath.Join("/tmp/", volName, "stage")
+		_, err = instance.SSH(fmt.Sprint("mkdir -p ", stageDir))
+
```
mattcary, line 62: Check the ssh error here?

mattcary: Actually, looking at the number of times we make directories and the cleanup we want to do after its done, maybe it would be better to at the beginning make a tmpdir (eg https://golang.org/pkg/io/ioutil/#TempDir), put all the staging & publishing directories in there, and then just remove the tmpdir at the end instead of removing all the intermediate directories each time.

Note the tmpdir might be necessary if tests are run in parallel on the same instance.

mattcary:
This looks good to me basically.

As Anna is going to be going adding much more tests you can defer the remote-tmp-dir change to her if you want, LMK.

saikat-royc:
> This looks good to me basically.
> 
> As Anna is going to be going adding much more tests you can defer the remote-tmp-dir change to her if you want, LMK.

Sounds good. The goal of this patch was to get the test running successfully. We can cleanup in subsequent patches.

PR #69, opened 2020-10-27
Find instance network dynamically

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

If no network is provided in the storage class, the driver controller can auto detect the network for the cluster where its running. We have seen kubetest --up by default use a non-default network to create GCE instances. With this patch, we do not need to manually provide the network name in the storage class.

Fixes #

Tested manually by deploying driver and a demo pod with PVC in both GCE cluster and GKE cluster.

```release-note
Find instance network dynamically
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ volume. Customizable parameters for volume creation include:
 | Parameter         | Values                  | Default                                | Description |
 | ---------------   | ----------------------- |-----------                             | ----------- |
 | tier              | "standard"<br>"premium" | "standard"                             | storage performance tier |
-| network           | string                  | "default"                              | VPC name |
+| network           | string                  | Cluster network auto-detected by the driver controller | VPC name |
 | reserved-ipv4-cidr| string		              | ""                                     | CIDR range to allocate Filestore IP Ranges from.<br>The CIDR must be large enough to accommodate multiple Filestore IP Ranges of /29 each |
 
 For Kubernetes clusters, these parameters are specified in the StorageClass.
--- a/pkg/cloud_provider/metadata/metadata.go
+++ b/pkg/cloud_provider/metadata/metadata.go
@@ -18,20 +18,28 @@ package metadata
 
 import (
 	"fmt"
+	"strings"
 
 	"cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
 )
 
 type Service interface {
 	GetZone() string
 	GetProject() string
+	GetNetwork() string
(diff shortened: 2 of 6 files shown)


pkg/cloud_provider/metadata/metadata.go
```diff
@@ -45,9 +53,15 @@ func NewMetadataService() (Service, error) {
 		return nil, fmt.Errorf("failed to get project: %v", err)
 	}
 
+	network, err := getNetwork()
```
msau42, line 56: If the controller runs outside of the user's project, is this going to work?

pkg/cloud_provider/metadata/metadata.go
```diff
@@ -58,3 +72,19 @@ func (manager *metadataServiceManager) GetZone() string {
...
+	return manager.network
+}
+
+func getNetwork() (string, error) {
+	s, err := metadata.Get(MetadataNetworkSuffix)
+	// network returned has the complete path e.g. projects/<Project-Id>/networks/<Network-Name>
+	if err == nil {
+		s = strings.TrimSpace(s)
+		s = s[strings.LastIndex(s, "/")+1:]
+		glog.Infof("Found network %v", s)
```
msau42, line 86: future cleanup to switch to klog

README.md
```diff
@@ -40,7 +40,7 @@ volume. Customizable parameters for volume creation include:
 | Parameter         | Values                  | Default                                | Description |
 | ---------------   | ----------------------- |-----------                             | ----------- |
 | tier              | "standard"<br>"premium" | "standard"                             | storage performance tier |
-| network           | string                  | "default"                              | VPC name |
+| network           | string                  | Cluster network auto-detected by the driver controller | VPC name |
```
msau42, line 43: Even if we can auto detect the network, the user needs to still create firewall rules right?

pkg/cloud_provider/metadata/metadata.go
mattcary: @saikat-royc I can create several projects in the mattcary.info org we can use to test out various network configurations. Let's sync offline on that.

saikat-royc:
Closing this patch, https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/71 enables explicitly setup of network name for CI runs.

PR #70, opened 2020-10-28
Add resize e2e tests

annapendleton:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

Commit 1: parallelize test setup - this change enables us to set up instances and drivers in parallel prior to running tests to minimize amount of setup time for each test as the test suite grow. Tests will access instances via a random test context. Drivers (and instances if delete-instances=true) will be cleaned up in parallel after execution of all tests.

Commit 2: Refactor tests to reuse same disk. Add online resize tests - tests ability for controller to expand filestore disk while disk is still mounted to a node.

Commit 3: add offline resize test - tests ability for controller to expand filestore disk while disk is not mounted to a node.

```release-note
Add resize e2e tests
```


annapendleton:
cc @saikat-royc

test/e2e/tests/setup_e2e_test.go
```diff
@@ -61,28 +63,49 @@ var _ = BeforeSuite(func() {
...
+			defer GinkgoRecover()
+			nodeID := fmt.Sprintf("gcfs-csi-e2e-%s", curZone)
+
+			i, err := remote.SetupInstance(*project, curZone, nodeID, *serviceAccount, computeService)
+			Expect(err).To(BeNil())
+
+			// Create new driver and client
+			testContext, err := testutils.GCFSClientAndDriverSetup(i)
+			Expect(err).To(BeNil(), "Set up new Driver and Client failed with error")
+			tcc <- testContext
```
saikat-royc, line 84: nice!

test/e2e/tests/resize_e2e_test.go
```diff
@@ -0,0 +1,112 @@
...
+			err = client.NodeUnpublishVolume(vol.GetVolumeId(), publishDir)
+			Expect(err).To(BeNil(), "NodeUnpublishVolume failed with error")
+		}()
+
+		// Verify pre-resized disk size
+		sizeTb, err := testutils.GetFSSizeInTb(instance, publishDir)
+		Expect(err).To(BeNil(), "Failed to get FSSize in TB")
+		Expect(sizeTb * util.Tb).To(Equal(minVolumeSize))
+
+		// Resize controller
```
saikat-royc, line 96: Few comments:
a) I think it will be good to create a function of this:
    1. resize api call 2. verification of resized instances.
That way we can add all cases in a single place and reuse in both online and offline test case.

b) Add resize call on the same instance with limit bytes set in the resize csi request. (it can be one simple case, where both required and limit bytes are set in the request. for exhaustive checks of limit and required bytes we can add to UTs if needed). Thoughts?
c) Add a resize call which increments in the order of bytes (because the doc says that https://cloud.google.com/filestore/docs/editing-instances#differences_in_scaling_behavior_by_service_tier basic tier can be incremented in the order of GBs)
d) Add a resize call with lesser than the current size of the instance. not sure what the API returns, but I expect it to fail. basic HDD and basic SSD can only be expanded.

All of resize calls, b c d can be done on the same instance.

saikat-royc:
Overall the test steps look good to me. You can add the timing logs in the client-wrappers.go in this patch or a separate patch that is fine.

test/e2e/tests/resize_e2e_test.go
annapendleton: Refactored tests to act on single instance. Added b/c/d to current implementation. Lmk your thoughts of current implementation w.r.t a).

annapendleton: Added timing log within client-wrapper in commit 1.

test/e2e/tests/resize_e2e_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	// Invalid resize to smaller amount
+	err = di.TestCtx.Client.ControllerExpandVolume(di.Volume.GetVolumeId(), minVolumeSize)
+	Expect(err).To(BeNil(), "Controller expand volume unexpected success for resize to invalid amount")
+	validateResizeDisk(newSizeBytes, di)
+}
+
+func offlineResizeDisk(di *DiskInfo) {
+	// Resize controller
+	var newSizeBytes int64 = int(minVolumeSize * 1.5)
```
saikat-royc, line 198: verify pull job complains about 1.5.
```vet: test/e2e/tests/resize_e2e_test.go:198:47: 1.5 (untyped float constant) truncated to int64```

File: test/e2e/tests/resize_e2e_test.go [hunk 4]
```diff
@@ -0,0 +1,242 @@
...
+			_, err = getDisk(diskInfo)
+			Expect(err).NotTo(BeNil(), "Could get deleted disk from cloud directly")
+		}()
+		validateDisk(diskInfo)
+
+		writeAndReadDisk(diskInfo)
+
+		offlineResizeDisk(diskInfo)
+
+		onlineResizeDisk(diskInfo)
```
saikat-royc, line 63: Looks clean now!

test/e2e/tests/resize_e2e_test.go
annapendleton: Fixed!

annapendleton:
Added test description in the validation for resize as well to help determine which test the validation fails on.

test/e2e/tests/resize_e2e_test.go
```diff
@@ -0,0 +1,241 @@
...
+	newSizeLimit = minVolumeSize * 3
+	err = di.TestCtx.Client.ControllerExpandVolumeWithLimit(di.Volume.GetVolumeId(), newSizeBytes, newSizeLimit)
+	Expect(err).To(BeNil(), "Controller expand volume failed for resize with valid limit")
+	validateResizeDisk(newSizeBytes, di, "online resize - resize with valid limit")
+
+	// Invalid resize to smaller amount
+	err = di.TestCtx.Client.ControllerExpandVolume(di.Volume.GetVolumeId(), minVolumeSize)
+	Expect(err).To(BeNil(), "Controller expand volume should not error for invalid resize to smaller amount.")
+	validateResizeDisk(newSizeBytes, di, "online resize - resize to smaller amount")
+}
```
mattcary, line 194: Do we want to test any writing to the disk to make sure it wasn't unmounted / reformatted during the resize?

test/e2e/tests/setup_e2e_test.go
```diff
@@ -61,28 +63,49 @@ var _ = BeforeSuite(func() {
...
-		}
+	for _, tc := range testContexts {
+		go func(curTC *remote.TestContext) {
+			defer GinkgoRecover()
+			err := remote.TeardownDriverAndClient(curTC)
+			Expect(err).To(BeNil(), "Teardown Driver and Client failed with error")
+			if *deleteInstances {
+				curTC.Instance.DeleteInstance()
+			}
+		}(tc)
```
mattcary, line 103: Don't you have to wait for the teardown to finish? main() at least won't wait for other goroutines before existing; I'm not sure if that's true in test or not.

mattcary (commented):
Just some nits

test/e2e/tests/resize_e2e_test.go
```diff
@@ -0,0 +1,241 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
saikat-royc, line 2: 2020

test/e2e/tests/resize_e2e_test.go
saikat-royc: +1 for offline resize test too. after resize, then stage/publish and test write

test/e2e/tests/resize_e2e_test.go
```diff
@@ -0,0 +1,241 @@
...
+	// Unmount Disk
+	err = di.TestCtx.Client.NodeUnpublishVolume(di.Volume.GetVolumeId(), secondPublishDir)
+	Expect(err).To(BeNil(), "NodeUnpublishVolume failed with error")
+}
+
+func onlineResizeDisk(di *DiskInfo) {
+	instance := di.TestCtx.Instance
+	volName := di.Name
+
+	// Stage Disk
```
saikat-royc, line 137: The step of stage(defer unstage), publish(defer unpublish) is a common theme in tests.
We can perhaps make a function StageAndMount() which returns a cleanup function. Saves duplicate code and keeps the code logic in one place to make any changes. Thoughts?

```
StageAndPublish() {
    // stage
    // publish
    return cleanup_func() {
      // unpublish 
      //. unstage
    }
}
```

now we can write tests
```
onlineresizetest() {
    cleanup = StageAndPublish()
    defer cleanup()
    // resize test cases.
}
```
```
writeAndReadDisk() {
    cleanup = StageAndPublish()
    defer cleanup()
    // write tests
    // Publish to second dir
    // defer unpublish()
    // more write tests
}
```

test/e2e/tests/resize_e2e_test.go
annapendleton: done!

test/e2e/tests/setup_e2e_test.go
annapendleton: I double checked the documentation and cannot find anything related to test being special with goroutines. I've added a waitgroup to account for this problem.

I also double checked the create logic and found that the create call currently can get stuck waiting forever if there is an error in the setup - I've changed it a little so that it will no longer hang, PTAL there as well!

test/e2e/tests/resize_e2e_test.go
annapendleton: Done - refactored the write / read as well into their own respective methods for easier testing.

test/e2e/tests/resize_e2e_test.go
annapendleton: I like this idea, implemented! 2 things worth adding:
1. previously I was risk averse to helper methods b/c default gingko doesn't print out full stack trace, so it was difficult to debug when errors occurred. I just discovered this new flag option "-trace" that solves this issue, so now it's much easier :) 

2. doing it this way ~slightly~ changes the test where the filestore instance is published to 2 directories simultaneously. I don't think this affects the test in any meaningful way.

mattcary:
lgtm. I'll let Saikat give the final ok.

PR #81, opened 2020-11-19
Fix the tag name for driver builds

saikat-royc:
When a commit is detected with a tag, the tag name should be used
instead of the commit sha.

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/build/generate_image_tags.sh
+++ b/build/generate_image_tags.sh
@@ -35,7 +35,7 @@ if [ $LATEST_GIT_TAG ]; then
 fi
 
 if [ $LATEST_TAG_REV ] && [ $LATEST_TAG_REV == $HEAD_REV ]; then
-  IMAGE_TAGS+=$LATEST_TAG_REV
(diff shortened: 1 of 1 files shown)


build/generate_image_tags.sh
```diff
@@ -35,7 +35,7 @@ if [ $LATEST_GIT_TAG ]; then
 fi
 
 if [ $LATEST_TAG_REV ] && [ $LATEST_TAG_REV == $HEAD_REV ]; then
```
mattcary, line 37: Do these LATEST_TAG_REV also need to change to the git tag?

saikat-royc: no that should be correct since it compares revision(or sha)

mattcary: Ah I see you compare with the hashes but need to mark with the tag. Thx

/lgtm

PR #85, opened 2020-11-20
Documentation and image bump to 0.3.0 for stable

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Documentation update and stable image bump to 0.3.0 before we cherry pick it to release-0.3 branch

Fixes #

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,10 @@ Disclaimer: This is not an officially supported Google product.
 This driver allows volumes backed by Google Cloud Filestore instances to be
 dynamically created and mounted by workloads.
 
-If multiple volumes are not needed, then Google Cloud Filestore instances can be
-manually created without this CSI driver and mounted using existing NFS volume
-plugins. Please see the Cloud Filestore
-[documentation](https://cloud.google.com/filestore/docs/accessing-fileshares)
-for more details.
-
 ## Project Status
-Status: Alpha
+Status: Beta
 
-Latest image: `gke.gcr.io/gcp-filestore-csi-driver:v0.2.0`
+Latest image: `gcr.io/k8s-staging-cloud-provider-gcp/gcp-filestore-csi-driver:v0.3.0`
 
 Also see [known issues](KNOWN_ISSUES.md) and [CHANGELOG](CHANGELOG.md).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,6 +29,7 @@ This plugin is compatible with CSI version 1.3.0.
 | Filestore CSI Driver\Kubernetes Version | 1.12 | 1.13 | 1.14 | 1.15 | 1.16+ |
```
mattcary, line 29: We can simplify the table by making a combined column 1.12 - 1.13 and just have a 1.15+ column

Although what versions are we testing 0.3.0 on? Maybe we should only advertise support for the versions where we're testing. k8s only supports 3 versions back anyway and since it's not going to matter for GKE maybe we should drop 1.14 support so we aren't on the hook for any bugs that come up.

There aren't any CRDs or anything that we depend on for older versions? The PD CSI GA version is only 1.17+ but I'm not sure why.

I guess topology is one thing that's changing (GA in 1.17), but this driver supports both that and the beta?

mattcary (commented):
I presume you're going to cut the 0.3.0 release after submitting this?

https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/releases

saikat-royc:
> I presume you're going to cut the 0.3.0 release after submitting this?
> 
> https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/releases

Yes

README.md
saikat-royc: 1.17 is due to provisioner topology, snapshot supports. I have updated the table

README.md
```diff
@@ -65,6 +60,11 @@ Note that non-default networks require extra [firewall setup](https://cloud.goog
   | Immediate            |       N/A         |        Present      | Call CreateVolume with requisite set to allowedTopology and preferred set to the sorted and shifted version of requisite at a randomized index |
   | Immediate            |       N/A         |        Not Present  | Call CreateVolume with requisite = aggregated topology across nodes which contain the topology keys of CSINode objects, preferred = sort and shift requisite at a randomized index |
 
+* Filestore Backups: Filestore Backup relies on CSI VolumeSnapshot which is a Beta feature in k8s enabled by default in
+Kubernetes 1.17+. CSI VolumeSnapshot should not be confused with Filestore Backups. Filestore CSI driver leverages CSI VolumeSnapshot capability to support Filestore Backups by specifying a `type` parameter in the VolumeSnapshotClass object. For more details see user-guide [here](docs/kubernetes/backup.md).
```
mattcary, line 64: Filestore snapshots is a publicly announced alpha feature: https://cloud.google.com/sdk/gcloud/reference/alpha/filestore/snapshots/create

I think the confusion is going to be with CSI snapshots vs filestore snapshots vs backups. 

Since the rest of this document is in terms of CSI/k8s features, maybe the bullet here should be from a CSI perspective instead of a filestore one. Namely, something like:

* Volume Snapshots: This driver currently supports CSI VolumeSnapshots on a Filestore instance using the GCP Filestore Backup feature. CSI VolumeSnapshot is a Beta feature in k8s enabled by default in 1.17+. The GCP [Filestore Snapshot alpha](https://cloud.google.com/sdk/gcloud/reference/alpha/filestore/snapshots/create) is not currently supported, but will be in the future via the `type` parameter in the VolumeSnapshotClass. For more details see the user-guide [here](docs/kubernetes/backup.md).

saikat-royc: Done

PR #98, opened 2021-01-13
k8s minor version overlays

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This patch enables creating overlays per k8s minor version. The overlays capture the manifest bundle for the driver for a given k8s version.

Fixes #

Summary of changes:
1. mapping of overlay to sidecar images
```
stable-1-16, prow-gke-release-staging-rc-1-16:
provisioner: 1.6.1
resizer: 1.1.0
node registrar: 1.3.0
filestore driver:0.3.1

stable-1-17, prow-gke-release-staging-rc-1-17:
stable-1-18, prow-gke-release-staging-rc-1-18:
stable-1-19, prow-gke-release-staging-rc-1-19:
provisioner: 2.1.0
resizer: 1.1.0
snapshotter:3.0.3
node registrar: 1.3.0
filestore driver:0.3.1

stable-master, prow-gke-release-staging-rc-master:
provisioner: 2.1.0
resizer: 1.1.0
snapshotter:4.0.0
node registrar: 1.3.0
filestore driver:0.3.1
```
2. Add test infra to run k8s integration test on GKE. This enables testing the various overlays on GKE and GCE.

TODO items in future patches:
1. add internal prow jobs to test the prow-gke-release-staging-rc* overlays and verify things work.
2. Remove the old prow prow-gke-release-staging-rc and stable overlays and the respective jobs, and point run-k8s-integration.sh to default stable-master overlay.
3. Update readme

```release-note
Split the overlays into per k8s minor version
```

### DIFF
--- a/deploy/kubernetes/images/prow-gke-release-staging-rc-1-19/image.yaml
+++ b/deploy/kubernetes/images/prow-gke-release-staging-rc-1-19/image.yaml
@@ -1,0 +1,45 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-provisioner-prow-rc
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-provisioner
+  newTag: "v2.1.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-resize-prow-rc
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-resizer
+  newTag: "v1.1.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-snapshotter-prow-rc
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-snapshotter
+  newTag: "v3.0.3"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-node-registrar-prow-rc
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-node-driver-registrar
+  newTag: "v1.3.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gce-fs-driver-rc
+imageTag:
+  name: k8s.gcr.io/cloud-provider-gcp/gcp-filestore-csi-driver
+  newName: gcr.io/k8s-staging-cloud-provider-gcp/gcp-filestore-csi-driver
+  newTag: "v0.3.1"
+---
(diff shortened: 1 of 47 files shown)


test/k8s-integration/main.go
```diff
@@ -355,6 +391,32 @@ func generateGCETestSkip(testParams *testParameters) string {
 	return skipString
 }
```
mattcary, line 392: I would have expected kustomize changes somewhere to use the new overlays. Or am I missing something?

deploy/kubernetes/images/prow-gke-release-staging-rc-1-19/image.yaml
```diff
@@ -0,0 +1,45 @@
...
+  newTag: "v1.1.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-snapshotter-prow-rc
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-snapshotter
+  newTag: "v3.0.3"
```
mattcary, line 25: The snapshotter is v4 in 1.19, do we not want to use that here?

saikat-royc: from the release page I see 4.0.0 has min 1.20 hence I didnot include in 1.19.
https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v4.0.0

mattcary:
Ah, you're right, the 0.4 component which is on 1.19 uses v3.0.2. My bad.
Component version vs driver version confusion strikes again.

test/k8s-integration/main.go
mattcary: From offline conversation: it's passed in via run-k8s-integration*.sh and which will appear in a separate PR.

verult (commented):
Overlays LGTM overall, thanks!

What about having stable/master, stable/1-19, stable/1-18, etc instead, i.e. nested directories? Would that break scripts?

if we make it stable/1-19 we would need some conversion logic to find the right directory for deploy or prow jobs.
for e.g using stable-1-19 directory name, we can directly pass 
```
- name: GCE_PD_OVERLAY_NAME
-value: stable-1-19
```
to the prow jobs with no additional code change.

deploy/kubernetes/overlays/stable-1-16/no_snapshotter_roles.yaml
```diff
@@ -0,0 +1,34 @@
+# TODO: Check if we can remove the snapshot, snapshotcontent get list roles for provisioner
+$patch: delete
```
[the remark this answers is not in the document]
saikat-royc: Thanks for bringing this up. I did give it some thought when making the changes. my reasoning is as follows:
I think it should be on-need basis. example adding to the current base is optimal when we are adding some component, which is going to be available in all future component. e.g adding a snapshotter sidecar- in this case deleting from 1.16 is more optimal (as its a one of deletion), else we would need to keep adding the snapshotter to all the future 1.17+ overlays.

If it's a one of change, say some change specific to a limited set of k8s version, addition to specifc k8s version overlay may be optimal.

PR #107, opened 2021-03-04
add cloud-config and some more fixes

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

1. Add support for cloud-config for GCE cloud provider (2bd88cd) to enable `--cloud-config=/etc/gce.conf ` args for managed GKE driver
2. Remove dependency on nodeId for controller service (c78f30c)
3. Minor fixes: a. kustomize version b. fix do-driver-build (there was a bug which did not replace the image tag) (4b18df2)

Fixes #

```release-note
Enable support for cloud provider config
```

### DIFF
--- a/pkg/cloud_provider/cloud.go
+++ b/pkg/cloud_provider/cloud.go
@@ -17,40 +17,159 @@ limitations under the License.
 package cloud
 
 import (
+	"context"
 	"fmt"
+	"net/http"
+	"os"
+	"time"
 
+	"cloud.google.com/go/compute/metadata"
+	gcpmeta "cloud.google.com/go/compute/metadata"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/compute/v1"
+	"gopkg.in/gcfg.v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/file"
-	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/metadata"
 )
 
 type Cloud struct {
-	File file.Service
-	Meta metadata.Service
+	File    file.Service
+	Project string
+	Zone    string
 }
 
-func NewCloud(version string) (*Cloud, error) {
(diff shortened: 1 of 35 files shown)


pkg/cloud_provider/cloud.go
```diff
@@ -17,40 +17,160 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/file"
-	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/metadata"
 )
 
 type Cloud struct {
 	File file.Service
-	Meta metadata.Service
+	//Meta    metadata.Service
```
mattcary, line 39: Remove commented out Meta?

mattcary (commented):
Maybe also add a line in the deployment documentation explaining how the different options work?

Also I'm not sure we need hostNetwork for the filestore driver---there may be enough of the required info from the WI metadata server (?). That's probably out of scope of this PR, though.

mattcary:
> Also I'm not sure we need hostNetwork for the filestore driver---there may be enough of the required info from the WI metadata server (?). That's probably out of scope of this PR, though.

It would be nice to document as well that hostNetwork means that WI can't be used to authorize the driver, the token source needs to be used.

saikat-royc:
> It would be nice to document as well that hostNetwork means that WI can't be used to authorize the driver, the token source needs to be used.

Ack. I ll add the doc changes in a separate PR, as this one is too big already

pkg/cloud_provider/cloud.go
saikat-royc: Done

saikat-royc:
> > Also I'm not sure we need hostNetwork for the filestore driver---there may be enough of the required info from the WI metadata server (?). That's probably out of scope of this PR, though.
> 
> It would be nice to document as well that hostNetwork means that WI can't be used to authorize the driver, the token source needs to be used.

Created an internal bug to investigate hostNetwork dependency and update the readme

PR #110, opened 2021-03-16
Clone master instead of pulling from archive

mattcary:
/kind bug

Same as https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/721

```release-note
NONE
```


saikat-royc:
We also can cherrypick to release-0.4 branch

PR #112, opened 2021-03-16
Fix driver image binary path in dockerfile COPY

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

In https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-gcp-filestore-push-images/1371882741854900224: make driver is building the binary in ```/home/prow/go/src/github.com/kubernetes-sigs/gcp-filestore-csi-driver```, but the final COPY command
Step 14/17 : COPY --from=builder /go/src/sigs.k8s.io/gcp-filestore-csi-driver/bin/${DRIVERBINARY} /${DRIVERBINARY}
COPY failed: stat /var/lib/docker/overlay2/7fea248a42aa55bc591ada90ab9b6fec0fca20f83bb6a8348c358ce2eb31c791/merged/go/src/sigs.k8s.io/gcp-filestore-csi-driver/bin/gcp-filestore-csi-driver: no such file or directory

is using a /go/src from the host

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ DRIVERBINARY=gcp-filestore-csi-driver
 
 $(info PULL_BASE_REF is $(PULL_BASE_REF))
 $(info GIT_TAG is $(GIT_TAG))
+$(info PWD is $(PWD))
 
 # A space-separated list of image tags under which the current build is to be pushed.
 # Determined dynamically.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -36,6 +37,11 @@ else
 endif
 $(info STAGINGIMAGE is $(STAGINGIMAGE))
 
+BIN_DIR=bin
```
BenTheElder, line 40: You can do this with just BINDIR?=bin https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors

saikat-royc: Done

PR #125, opened 2021-06-10
Serialize volume operations for a given volume and return appropriate error codes

saikat-royc:
1. Use mutex for volume ops locking
2. Use appropriate error codes for expected long running ops

**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Filestore operations are long running by default. We need to differentiate expected operations from internal errors to avoid false positives in SLO breach. This PR ensures that for expected filestore operations DeadlineExceeded is returned.
Further, volume locking is implemented in order to serialize volume operations in controller and node.
- For a given volume ID, create/delete/createsnapshot is serialized.
- For a given volume ID, node stage/unstage operation is serialized
- Node publish/unpublish operations are serialized at a targetPath key, since we want to node publish operations to work in parallel, for a given staged volume.

Fixes #

Test manually by reducing the caller timeout flags of csi-provisioner and csi-snapshotter sidecar [this](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/deploy/kubernetes/base/controller/controller.yaml#L30) and [this](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/deploy/kubernetes/base/controller/controller.yaml#L90) to very small value like 20 seconds, and checked that appropriate deadline exceeded errors are returned during filestore instance provision/delete and create/delete snapshot ops.
Also tested with deployment of ~50 replicas scheduled on a single node, to ensure node publish work in parallel for each pod.

```release-note
Serialize volume operations for a given volume and return appropriate error codes
```

### DIFF
--- a/pkg/cloud_provider/file/fake.go
+++ b/pkg/cloud_provider/file/fake.go
@@ -198,3 +198,33 @@ func notFoundError() *googleapi.Error {
 		},
 	}
 }
+
+type fakeBlockingServiceManager struct {
+	*fakeServiceManager
+	// 'OperationUnblocker' channel is used to block the execution of the respective function using it. This is done by sending a channel of empty struct over 'OperationUnblocker' channel, and wait until the tester gives a go-ahead to proceed further in the execution of the function.
+	OperationUnblocker chan chan struct{}
+}
+
+func NewFakeBlockingService(operationUnblocker chan chan struct{}) (Service, error) {
+	return &fakeBlockingServiceManager{
+		fakeServiceManager: &fakeServiceManager{
+			createdInstances: map[string]*ServiceInstance{},
+			backups:          map[string]*BackupInfo{},
+		},
+		OperationUnblocker: operationUnblocker,
+	}, nil
+}
+
+func (m *fakeBlockingServiceManager) CreateInstance(ctx context.Context, obj *ServiceInstance) (*ServiceInstance, error) {
+	execute := make(chan struct{})
+	m.OperationUnblocker <- execute
+	<-execute
+	return m.fakeServiceManager.CreateInstance(ctx, obj)
+}
+
+func (m *fakeBlockingServiceManager) DeleteInstance(ctx context.Context, obj *ServiceInstance) error {
+	execute := make(chan struct{})
+	m.OperationUnblocker <- execute
+	<-execute
+	return m.fakeServiceManager.DeleteInstance(ctx, obj)
+}
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -81,6 +81,7 @@ type controllerServerConfig struct {
 	fileService file.Service
 	cloud       *cloud.Cloud
 	ipAllocator *util.IPAllocator
+	volumeLocks *util.VolumeLocks
 }
 
 func newControllerServer(config *controllerServerConfig) csi.ControllerServer {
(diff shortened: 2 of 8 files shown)


pkg/csi_driver/controller.go
```diff
@@ -136,6 +137,12 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	volumeID := getVolumeIDFromFileInstance(newFiler, modeInstance)
```
mattcary, line 140: Why are we doing this after checking for a snapshot source? I'd think we'd want to block before collecting such info so that there's less of a window for a race, eg a snapshot getting deleted.

pkg/csi_driver/controller_test.go
```diff
@@ -52,6 +52,7 @@ func initTestController(t *testing.T) csi.ControllerServer {
 		driver:      initTestDriver(t),
 		fileService: fileService,
 		cloud:       cloudProvider,
+		volumeLocks: util.NewVolumeLocks(),
```
mattcary, line 55: We should be able to add some simple tests to run through the volume locking code paths, no?

saikat-royc: Ack

pkg/csi_driver/controller.go
```diff
@@ -229,14 +241,32 @@ func (s *controllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolu
...
 	}
+
 	filer, _, err := getFileInstanceFromID(volumeID)
 	if err != nil {
 		// An invalid ID should be treated as doesn't exist
 		glog.V(5).Infof("failed to get instance for volume %v deletion: %v", volumeID, err)
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
+	if acquired := s.config.volumeLocks.TryAcquire(volumeID); !acquired {
```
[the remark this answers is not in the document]
saikat-royc: Done

pkg/csi_driver/controller.go
saikat-royc: Moved it

pkg/cloud_provider/file/fake.go
```diff
@@ -198,3 +198,32 @@ func notFoundError() *googleapi.Error {
 		},
 	}
 }
+
+type fakeBlockingServiceManager struct {
```
mattcary, line 202: Can you document this a bit? There's two non-obvious things going on here (1) a channel of channels (readyToExecute), and (2) why readyToExecute in the test below needs to have a buffer of 1.

saikat-royc: Done

pkg/csi_driver/controller_test.go
```diff
@@ -52,6 +55,25 @@ func initTestController(t *testing.T) csi.ControllerServer {
...
+	})
+}
+
+func initBlockingTestController(t *testing.T, readyToExecute chan chan struct{}) csi.ControllerServer {
```
mattcary, line 62: s/readyToExecute/operationUnblocker/ in this method.

saikat-royc: Oops. Fixed

PR #143, opened 2021-07-08
cleaning up the clusterUpGKE

leiyiz:
**What type of PR is this?**
/kind cleanup

replace hard-coded options with options and env vars

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -292,26 +292,44 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 			return fmt.Errorf("failed to create a new service: %v", err)
 		}
 
-		// TODO: change to read CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER env var
-		service.BasePath = "https://test-container.sandbox.googleapis.com/" // e.g test, staging env URL
+		service.BasePath = os.Getenv("CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER")
+		if service.BasePath == "" {
+			service.BasePath = "https://container.googleapis.com/"
+		}
 
-		// TODO: honor the gkeNodeVersion flag
+		klog.Infof("targeting api endpoint: %s", service.BasePath)
+
 		request := &container.CreateClusterRequest{
 			Cluster: &container.Cluster{
-				Name:             *gkeTestClusterName,
-				InitialNodeCount: int64(numNodes),
+				Name: *gkeTestClusterName,
 				AddonsConfig: &container.AddonsConfig{
 					GcpFilestoreCsiDriverConfig: &container.GcpFilestoreCsiDriverConfig{Enabled: true},
 				},
-				ReleaseChannel: &container.ReleaseChannel{Channel: "RAPID"},
-				NodeConfig: &container.NodeConfig{
-					MachineType: "n1-standard-2",
-					ImageType:   imageType,
-					OauthScopes: []string{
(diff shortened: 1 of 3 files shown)


leiyiz:
@saikat-royc

test/k8s-integration/cluster.go
```diff
@@ -292,28 +292,42 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 			return fmt.Errorf("failed to create a new service: %v", err)
 		}
 
-		// TODO: change to read CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER env var
-		service.BasePath = "https://test-container.sandbox.googleapis.com/" // e.g test, staging env URL
+		service.BasePath = os.Getenv("CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER")
```
saikat-royc, line 295: note: for prod, the env variable will be unset. All though we only have test jobs for test and staging env. If we want to run jobs for prod, this maynot work.

leiyiz: oh I see... what's the endpoint for prod? I think a if statement here would be fine

saikat-royc: it must be "container.sandbox.googleapis.com" i think. can you check in your setup?

saikat-royc:
@leiyiz can you check the logic of generating skip string and handle node version skew (i.e when master and node pool are at different versions)? https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L549

saikat-royc:
overall changes look good. one nit: can you print the gcloud cluster list output right before waitForNodeDaemonset() so that we know exactly what cluster version was provisioned

test/k8s-integration/main.go
```diff
@@ -424,27 +422,23 @@ func generateGCETestSkip(testParams *testParameters) string {
...
+		nodeVer = mustParseVersion(testParams.nodeVersion)
 	}
 
 	// Skip fsgroup change policy based on GKE cluster version.
-	if curVer.lessThan(mustParseVersion("1.20.0")) {
+	if curVer.lessThan(mustParseVersion("1.20.0")) || (nodeVer != nil && nodeVer.lessThan(mustParseVersion("1.20.0"))) {
 		skipString = skipString + "|fsgroupchangepolicy"
 	}
 
 	// Running snapshot tests on GKE needs k8s 1.19x storage e2e testsuite (until GKE supports v1 snapshot APIs).
```
saikat-royc, line 435: Running snapshot tests on GKE clusters which does not support v1 snapshot CRDs needs 1.19x storage e2e testsuites...

test/k8s-integration/main.go
```diff
@@ -424,27 +422,23 @@ func generateGCETestSkip(testParams *testParameters) string {
...
-	if curVer.lessThan(mustParseVersion("1.16.0")) {
-		skipString = skipString + "|allowExpansion"
+	var nodeVer *version
+	if testParams.nodeVersion != "" {
+		nodeVer = mustParseVersion(testParams.nodeVersion)
 	}
 
 	// Skip fsgroup change policy based on GKE cluster version.
-	if curVer.lessThan(mustParseVersion("1.20.0")) {
+	if curVer.lessThan(mustParseVersion("1.20.0")) || (nodeVer != nil && nodeVer.lessThan(mustParseVersion("1.20.0"))) {
```
saikat-royc, line 431: we need a similar handling on 1.x.y-gke.z for useStagingDriver here?

PR #144, opened 2021-07-13
fixed issue with unability to recognize gce-region flag is set

leiyiz:
**What type of PR is this?**
/kind failing-test

test failure in prow

test failure [here](https://prow-gob.gcpnode.com/view/gs/gob-prow/logs/periodic-gke-pd-csi-driver-k8s-integration-preview-regional-test-env/1414991095246360576) caused by test script's inability to recognize `gce-region` flag being set

```release-note
None
```


leiyiz:
@saikat-royc

PR #147, opened 2021-07-21
update v1beta1 api and move filestore api calls to beta1 api

leiyiz:
**What type of PR is this?**
 /kind api-change

we are moving to beta1 api because of upcoming Filestore new enterprise tier support.  Using the v1beta1 for all API calls, simplifies the code and will enable provisioning of all tiers basic, premium, high scale and enterprise.

```release-note
Update Filestore APIs to use v1beta1
```


leiyiz:
@saikat-royc

PR #164, opened 2021-08-06
remove the hard-coded 29 for ipRangeSize

leiyiz:
**What type of PR is this?**
 /kind bug

cidr range size was hardcoded, and enterprise tier filestore needs 24 instead of 29

**Special notes for your reviewer:**
manual tests by providing `reserved-ipv4-cidr` in a enterprise tier filestore csi storageClass and verified that the provision succeeded.

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -171,14 +171,14 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 			return nil, status.Error(codes.Internal, msg)
 		}
 	} else {
-		// If we are creating a new instance, we need pick an unused /29 range from reserved-ipv4-cidr
+		// If we are creating a new instance, we need pick an unused CIDR range from reserved-ipv4-cidr
 		// If the param was not provided, we default reservedIPRange to "" and cloud provider takes care of the allocation
 		if reservedIPV4CIDR, ok := req.GetParameters()[paramReservedIPV4CIDR]; ok {
 			reservedIPRange, err := s.reserveIPRange(ctx, newFiler, reservedIPV4CIDR)
 
 			// Possible cases are 1) CreateInstanceAborted, 2)CreateInstance running in background
 			// The ListInstances response will contain the reservedIPRange if the operation was started
-			// In case of abort, the /29 IP is released and available for reservation
+			// In case of abort, the CIDR IP is released and available for reservation
 			defer s.config.ipAllocator.ReleaseIPRange(reservedIPRange)
 			if err != nil {
 				return nil, err
@@ -218,7 +218,11 @@ func (s *controllerServer) reserveIPRange(ctx context.Context, filer *file.Servi
 	if err != nil {
 		return "", err
 	}
-	unreservedIPBlock, err := s.config.ipAllocator.GetUnreservedIPRange(cidr, cloudInstancesReservedIPRanges)
+	ipRangeSize := util.IpRangeSize
+	if filer.Tier == enterpriseTier {
+		ipRangeSize = util.IpRangeSizeEnterprise
+	}
+	unreservedIPBlock, err := s.config.ipAllocator.GetUnreservedIPRange(cidr, ipRangeSize, cloudInstancesReservedIPRanges)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/util/ip_def.go
+++ b/pkg/util/ip_def.go
@@ -1,0 +1,25 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+const (
+	// the ipRangeSize for Filestore instances
+	IpRangeSize = 29
+
+	// the ipRangeSize for enterprise tier Filestore instances
+	IpRangeSizeEnterprise = 24
+)
--- a/pkg/util/ip_reservation.go
+++ b/pkg/util/ip_reservation.go
@@ -24,22 +24,11 @@ import (
 )
 
 const (
-	// Size of the network address of the IPRange we intend to reserve
-	ipRangeSize = 29
-	// Maximum value of a byte
-	byteMax = 255
 	// Total number of bits in an IPV4 address
 	ipV4Bits = 32
 )
 
-var (
-	// step size for IP range increment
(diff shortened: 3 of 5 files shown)


pkg/util/ip_reservation.go
```diff
@@ -34,9 +32,9 @@ const (
 
 var (
 	// step size for IP range increment
-	incrementStep29IPRange = (byte)(math.Exp2(ipV4Bits - ipRangeSize))
```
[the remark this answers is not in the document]
leiyiz: done.

leiyiz:
@saikat-royc

pkg/util/ip_reservation.go
```diff
@@ -103,7 +94,7 @@ func (ipAllocator *IPAllocator) GetUnreservedIPRange(cidr string, cloudInstances
 		reservedIPRanges[reservedIPRange] = true
 	}
 
-	for cidrIP := cloneIP(ip.Mask(ipnet.Mask)); ipnet.Contains(cidrIP) && err == nil; cidrIP, err = incrementIP(cidrIP, incrementStep29IPRange) {
+	for cidrIP := cloneIP(ip.Mask(ipnet.Mask)); ipnet.Contains(cidrIP) && err == nil; cidrIP, err = incrementIP(cidrIP, (byte)(math.Exp2(float64(ipV4Bits-ipRangeSize)))) {
```
mattcary, line 97: To make it self-commenting, pull out incrementStepIPRange = (byte)(math.Exp2(float64(ipV4Bits - ipRangeSize))

pkg/util/ip_reservation_test.go
```diff
@@ -212,7 +215,7 @@ func getIPRanges(cidr string, ipRangesCount int, t *testing.T) map[string]bool {
 	// 1) We have the required number of IP ranges in the set
 	// 2) IP range overflow occurs and IP increment is not possible
 	// 3) The incremented IP range is not contained in the cidr
-	for cidrIP := ip.Mask(ipRangeMask); ipnet.Contains(cidrIP) && err == nil && i < ipRangesCount; cidrIP, err = incrementIP(cidrIP, incrementStep29IPRange) {
+	for cidrIP := ip.Mask(ipRangeMask); ipnet.Contains(cidrIP) && err == nil && i < ipRangesCount; cidrIP, err = incrementIP(cidrIP, (byte)(math.Exp2(float64(ipV4Bits-ipRangeSize)))) {
```
mattcary, line 218: and here

mattcary:
after squashing the commits this will lgtm

pkg/csi_driver/controller.go
```diff
@@ -218,7 +218,11 @@ func (s *controllerServer) reserveIPRange(ctx context.Context, filer *file.Servi
 	if err != nil {
 		return "", err
 	}
-	unreservedIPBlock, err := s.config.ipAllocator.GetUnreservedIPRange(cidr, cloudInstancesReservedIPRanges)
+	ipRangeSize := 29
```
saikat-royc, line 221: create constants for these numbers 29 and 24

pkg/util/ip_def.go
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
saikat-royc, line 2: 2021

pkg/util/ip_reservation_test.go
```diff
@@ -89,6 +90,7 @@ func TestParseCIDR(t *testing.T) {
 func TestGetUnReservedIPRange(t *testing.T) {
 	// Using IPs which are not the beginning IPs of /29 CIDRs to evaluate the edge case
 	ips := [8]string{"192.168.92.3/29", "192.168.92.10/29", "192.168.92.20/29", "192.168.92.28/29"}
+	ipRangeSize := 29
```
saikat-royc, line 93: use the constant that was defined in ip_defs?

Lets add test case for 24 also.

pkg/util/ip_reservation.go
```diff
@@ -162,36 +153,28 @@ func (ipAllocator *IPAllocator) parseCIDR(cidr string) (net.IP, *net.IPNet, erro
 	}
 
 	// The IP specified in the reserved-ipv4-cidr must be aligned on the /29 network boundary
-	if ip.String() != ip.Mask(ipRangeMask).String() {
+	if ip.String() != ip.Mask(net.CIDRMask(ipRangeSize, ipV4Bits)).String() {
```
saikat-royc, line 156: change the statement to generic
```
the IP specified in the reserved-ipv4-cidr must be aligned on the /29 network boundary
```

pkg/util/ip_reservation_test.go
```diff
@@ -116,8 +130,42 @@ func TestGetUnReservedIPRange(t *testing.T) {
 			errorExpected: false,
 		},
 		{
-			name:            "1 Pending 0 Used",
+			name:            "0 Pending, 1 /24 Used enterprise",
+			cidr:            "192.168.92.0/22",
+			ipRangeSize:     IpRangeSizeEnterprise,
+			pendingIPRanges: make(map[string]bool),
+			cloudProviderReservedIPRanges: map[string]bool{
+				ipsLarge[0]: true,
```
saikat-royc, line 138: I think we can directly use the string values here. makes the test case more readable instead of looking to find the value for the key. Thoughts?

pkg/util/ip_reservation_test.go
```diff
@@ -126,8 +174,42 @@ func TestGetUnReservedIPRange(t *testing.T) {
...
+			ipRangeSize:     IpRangeSize,
+			pendingIPRanges: make(map[string]bool),
+			cloudProviderReservedIPRanges: map[string]bool{
+				ipsLarge[0]: true,
+			},
+			expected:      "192.168.93.0/29",
+			errorExpected: false,
+		},
+		{
+			name:            "1 /24 Pending 0 Used enterprise",
```
saikat-royc, line 199: test case name says 1 /24 pending, but `cloudProviderReservedIPRanges` has the range range added. Is this correct? same with the other non zero pending test cases.

saikat-royc:
we also need to handle pagination of ListInstances() call.

mattcary:
> we also need to handle pagination of ListInstances() call.

That seems orthogonal? Might be cleaner to do in a separate PR.

pkg/util/ip_reservation_test.go
leiyiz: /27 has 4 /29 ranges, and ips[i] is in the ith range, so it's a lot easier to think about once understand that. the same goes with ipsLarge. so I think it'd be good to keep them as is. Maybe add some comments to explain that part.

pkg/util/ip_reservation.go
leiyiz: sure

saikat-royc:
> > we also need to handle pagination of ListInstances() call.
> 
> That seems orthogonal? Might be cleaner to do in a separate PR.

If the list instance is not handled correctly, the cloudInstancesReservedIPRanges will not be setup correctly. But yes, we can take it up as a separate PR.

pkg/util/ip_reservation_test.go
saikat-royc: Ack,. add more comments

saikat-royc:
/lgtm
nice work!

leiyiz:
@saikat-royc

PR #178, opened 2021-09-13
new storageclass and increase pvc timeout for enterprise tier prow test

leiyiz:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

In order to run prow tests with enterprise tier storage-class, we need to have a storage-class and increase the provisioning timeout because enterprise tier pv takes ~20 min to provision.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/config/fs-sc-enterprise-default-network.yaml
+++ b/test/k8s-integration/config/fs-sc-enterprise-default-network.yaml
@@ -1,0 +1,8 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-filestore-enterprise
+provisioner: filestore.csi.storage.gke.io
+parameters:
+  tier: enterprise
+volumeBindingMode: WaitForFirstConsumer
--- a/test/k8s-integration/driver-config.go
+++ b/test/k8s-integration/driver-config.go
@@ -26,10 +26,12 @@ const (
 	testConfigDir      = "test/k8s-integration/config"
 	configTemplateFile = "test-config-template.in"
 	configFile         = "test-config.yaml"
+	enterpriseTier     = "enterprise"
 
 	// configurable timeouts for the k8s e2e testsuites.
-	podStartTimeout       = "600s"
-	claimProvisionTimeout = "600s"
+	podStartTimeout                 = "600s"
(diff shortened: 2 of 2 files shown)


saikat-royc:
We also need to use specific boskos project with the quota enabled for the enterprise tier. We also need to check how many maximum tests run in parallel, that would be approximate indicator of how many 1T enterprise tier can be provisioned in parallel. I think a quota of 10T would be sufficient to start with.

saikat-royc:
we then need to add the newly created SC [here](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/test/k8s-integration/main.go#L49)

leiyiz:
I believe the boskos projects and the sc files are specified not here but in config_gen/ on Gerrit.

leiyiz:
> We also need to check how many maximum tests run in parallel

What would you say is a good way to do that?

test/k8s-integration/config/fs-sc-enterprise-default-network.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-filestore
```
saikat-royc, line 4: csi-filestore-enterprise

test/k8s-integration/driver-config.go
```diff
@@ -29,7 +29,7 @@ const (
 
 	// configurable timeouts for the k8s e2e testsuites.
 	podStartTimeout       = "600s"
-	claimProvisionTimeout = "600s"
+	claimProvisionTimeout = "1800s"
```
saikat-royc, line 32: lets use timeouts per tier. large timeout can hide issues in standard and premium tiers

leiyiz: done

test/k8s-integration/config/fs-sc-enterprise-default-network.yaml
leiyiz: done

leiyiz:
@saikat-royc can you ok-to-test this

PR #192, opened 2021-12-30
enable dynamic provision with shared-vpc from service project

leiyiz:
in order to do that, we need to pass appropriate `connect-mode`

**What type of PR is this?**

/kind feature

provision Filestore instances with shared-vpc from service project went into public preview and we need to support it in filestore csi driver

Fixes #

```release-note
new parameter `connect-mode` needs to be set to `PRIVATE_SERVICE_ACCESS` if provisioning Filestore instances from service projects of shared-vpc. If a specific range needs to be specified, `reserved-ip-range` needs to be used instead of a CIDR range
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -40,6 +40,9 @@ const (
 	enterpriseTier = "enterprise"
 	defaultNetwork = "default"
 
+	directPeering        = "DIRECT_PEERING"
+	privateServiceAccess = "PRIVATE_SERVICE_ACCESS"
+
 	// Keys for Topology.
 	TopologyKeyZone = "topology.gke.io/zone"
 )
@@ -56,6 +59,8 @@ const (
 	paramLocation         = "location"
 	paramNetwork          = "network"
 	paramReservedIPV4CIDR = "reserved-ipv4-cidr"
+	paramReservedIPRange  = "reserved-ip-range"
+	paramConnectMode      = "connect-mode"
 
 	// Keys for PV and PVC parameters as reported by external-provisioner
 	ParameterKeyPVCName      = "csi.storage.k8s.io/pvc/name"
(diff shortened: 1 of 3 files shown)


pkg/csi_driver/controller.go
```diff
@@ -394,9 +405,15 @@ func (s *controllerServer) generateNewFileInstance(name string, capBytes int64,
 			}
 		case paramNetwork:
 			network = v
+		case paramConnectMode:
+			connectMode = v
+			if connectMode != defaultConnectMode && connectMode != privateConnectMode {
+				return nil, fmt.Errorf("connect mode can only be one of %v or %v", defaultConnectMode, privateConnectMode)
+			}
```
jingxu97, line 412: use %q instead of %v?

pkg/csi_driver/controller.go
```diff
@@ -40,6 +40,9 @@ const (
 	enterpriseTier = "enterprise"
 	defaultNetwork = "default"
 
+	defaultConnectMode = "DIRECT_PEERING"
+	privateConnectMode = "PRIVATE_SERVICE_ACCESS"
```
saikat-royc, line 44: privateServiceAccess

pkg/csi_driver/controller.go
```diff
@@ -40,6 +40,9 @@ const (
 	enterpriseTier = "enterprise"
 	defaultNetwork = "default"
 
+	defaultConnectMode = "DIRECT_PEERING"
```
saikat-royc, line 43: directPeering

pkg/csi_driver/controller.go
```diff
@@ -394,9 +405,15 @@ func (s *controllerServer) generateNewFileInstance(name string, capBytes int64,
 			}
 		case paramNetwork:
 			network = v
+		case paramConnectMode:
+			connectMode = v
+			if connectMode != defaultConnectMode && connectMode != privateConnectMode {
+				return nil, fmt.Errorf("connect mode can only be one of %v or %v", defaultConnectMode, privateConnectMode)
```
saikat-royc, line 411: there is a "CONNECT_MODE_UNSPECIFIED". Is this a default value? https://cloud.google.com/filestore/docs/reference/rest/v1/projects.locations.instances#ConnectMode

leiyiz: According to the same page it says "If not provided, the connect mode defaults to DIRECT_PEERING."

saikat-royc: so we should also support CONNECT_MODE_UNSPECIFIED as an acceptable string ?

leiyiz: No we should not, checked with Filestore team and the "FOO_UNSPECIFIED" is just a proto enum specificity

pkg/csi_driver/controller.go
leiyiz: wouldn't the `defaultConnectMode` be more aligned with the `defaultNetwork` above

pkg/csi_driver/controller.go
leiyiz: ditto above

pkg/csi_driver/controller.go
leiyiz: done

pkg/csi_driver/controller_test.go
```diff
@@ -456,7 +459,48 @@ func TestGenerateNewFileInstance(t *testing.T) {
...
+					ConnectMode: directPeering,
+				},
+				Volume: file.Volume{
+					Name:      newInstanceVolume,
+					SizeBytes: testBytes,
+				},
+			},
+		},
+		{
+			name: "custom params, private connect mode",
```
saikat-royc, line 472: can you also add invalid param test case for the invalid connectMode

leiyiz: done

PR #203, opened 2022-01-25
return InvalidArgument for CreateInstance failures

amacaskill:
**What type of PR is this?**
> /kind failing-test

This PR distinguishes the user errors from the other internal errors that are returned by CreateInstance. Right now, we are being alerted for CreateInstance failures that are user errors, and we don't want to include these in the calculation of the measured happiness for the filestore pod set up success rate SLO. For example, we see a lot of errors like 
` GRPC error: rpc error: code = Internal desc = CreateInstance operation failed: googleapi: Error 429: Quota limit 'StandardStorageGbPerRegion' has been exceeded. Limit: 4096 in region us-central1. ` and 
`
GRPC error: rpc error: code = Internal desc = googleapi: Error 403: Cloud Filestore API has not been used in project xxxxx before or it is disabled.
` These are both user errors that will now have code InvalidArgument instead of Internal. 
 Internal error codes are counted as failures, but InvalidArgument error codes are counted as successes, so this changes user errors to InvalidArgument code. 

```release-note
Users will now see the InvalidArgument error code for the 403 and 429 googleapi errors. 
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -214,7 +214,11 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		}
 		if createErr != nil {
 			glog.Errorf("Create volume for volume Id %s failed: %v", volumeID, createErr)
-			return nil, status.Error(codes.Internal, createErr.Error())
+			if file.IsUserError(createErr) {
+				return nil, status.Error(codes.InvalidArgument, createErr.Error())
+			} else {
+				return nil, status.Error(codes.Internal, createErr.Error())
+			}
 		}
 	}
 	resp := &csi.CreateVolumeResponse{Volume: fileInstanceToCSIVolume(filer, modeInstance, sourceSnapshotId)}
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -29,6 +29,7 @@ import (
 	"github.com/golang/glog"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
+	"google.golang.org/genproto/googleapis/rpc/code"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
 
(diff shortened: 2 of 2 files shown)


pkg/csi_driver/controller.go
```diff
@@ -214,7 +214,14 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		}
 		if createErr != nil {
 			glog.Errorf("Create volume for volume Id %s failed: %v", volumeID, createErr)
-			return nil, status.Error(codes.Internal, createErr.Error())
+			// Return InvalidArgument code for googleapi errors that are caused by users such as Error 429: Quota limit exceeded
```
mattcary, line 217: nit: move this comment to the IsUserError() ftn definition (eg, if we called this ftn a second time from somewhere else we wouldn't want to copy the comment)

pkg/csi_driver/controller.go
```diff
@@ -214,7 +214,14 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
...
+			// and Error 403: Permission Denied.
+			if file.IsUserError(createErr) {
```
mattcary, line 219: I wonder if it would be easier to just do return nil,  FilestoreAPIError(createErr), and then have FilestoreApiError map filestore API error codes to grpc error codes. That way if it ever gets more complicated we have all the error logic in the same place.

@saikat-royc ?

amacaskill: Right now that seems a little difficult for the IsNotFoundErr function. (I am assuming we would want to do this for both the IsNotFoundErr and the IsUserErr functions). There are 6 if checks for  `if file.IsNotFoundErr(err)` and it doesn't always return the same grpc error code. Sometimes it [returns codes.NotFound](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/controller.go#L144-L147),  sometimes it [doesn't return an error](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/controller.go#L279-L281), some times it returns codes.Internal. So sometimes in different contexts a googleapi not found error would mean you would want to return a different  grpc error code, or none at all.

mattcary: Good point, makes sense to me.

PR #205, opened 2022-02-03
build driver image using docker buildx instead of docker build

leiyiz:
**What type of PR is this?**
/kind feature

In order to support multi-arch build for Filestore CSI driver, we need to build the driver image with docker buildx

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,11 +85,17 @@ push-webhook-image: webhook-image
 	}
 
 # Build the docker image for the core CSI driver.
-image:
+image: init-buildx
 		{                                                                   \
 		set -e ;                                                            \
-		for i in $(STAGINGVERSION) ;                                        \
-			do docker build --build-arg DRIVERBINARY=$(DRIVERBINARY) --build-arg TAG=$(STAGINGVERSION) -t $(STAGINGIMAGE):$${i} .; \
+		for i in $(STAGINGVERSION) ; do                                     \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -85,11 +85,20 @@ push-webhook-image: webhook-image
...
-			do docker build --build-arg DRIVERBINARY=$(DRIVERBINARY) --build-arg TAG=$(STAGINGVERSION) -t $(STAGINGIMAGE):$${i} .; \
+		for i in $(STAGINGVERSION) ; do                                     \
+			docker buildx build \
+				--platform linux/amd64 \
+				--build-arg TAG=$(STAGINGVERSION) \
+				--build-arg VERSION=$(STAGINGVERSION) \
+				--build-arg STAGINGVERSION=$(STAGINGVERSION) \
+				--build-arg BUILDPLATFORM=linux/amd64 \
+				--build-arg TARGETPLATFORM=linux/amd64 \
+				--build-arg ARCH=amd64 \
```
mattcary, line 99: It looks like ARCH, TAG, and VERSION aren't actually used as build args in the Dockerfile? Should probably remove them to keep things clean.

We don't use them in the pdcsi driver any more either, probably build-image in louhi should be cleaned up but that's a separate task.

leiyiz: DONE.

PR #229, opened 2022-02-25
create volume changes

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Layout the code for CSI create volume for multishare instances.
1. Multishare controller handles the CSI Create Volume call.
2. Multishare Ops manager handles cache consistency, manages lifecycle of instance and share ops.
3. cloud provider providers skeleton code for multishare APIs, and multi-share instance and share definitions.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -42,25 +42,32 @@ type PollOpts struct {
 }
 
 type Share struct {
-	Name      string
-	SizeBytes int64
+	Name           string              // only the share name
+	Parent         *MultishareInstance // parent captures the project, location details.
+	State          string
+	MountPointName string
+	Labels         map[string]string
+	CapacityBytes  int64
 }
 
 type MultishareInstance struct {
-	Project  string
-	Name     string
-	Location string
-	Tier     string
-	Network  Network
-	Shares   []Share
-	Labels   map[string]string
-	State    string
+	Project            string
+	Name               string
+	Location           string
+	Tier               string
+	Network            Network
+	CapacityBytes      int64
+	MaxCapacityBytes   int64
+	CapacityStepSizeGb int64
+	Labels             map[string]string
+	State              string
+	KmsKeyName         string
 }
 
 type ListFilter struct {
-	Project  string
-	Location string
-	Instance string
+	Project      string
+	Location     string
+	InstanceName string
 }
 
 type ServiceInstance struct {
@@ -104,15 +111,19 @@ type Service interface {
 	CreateInstanceFromBackupSource(ctx context.Context, obj *ServiceInstance, volumeSourceSnapshotId string) (*ServiceInstance, error)
 	HasOperations(ctx context.Context, obj *ServiceInstance, operationType string, done bool) (bool, error)
 	// Multishare ops
-	GetMultishareInstance(ctx context.Context, uri string) (*MultishareInstance, error)
+	GetMultishareInstance(ctx context.Context, obj *MultishareInstance) (*MultishareInstance, error)
 	ListMultishareInstances(ctx context.Context, filter *ListFilter) ([]*MultishareInstance, error)
 	StartCreateMultishareInstanceOp(ctx context.Context, obj *MultishareInstance) (*filev1beta1.Operation, error)
 	StartDeleteMultishareInstanceOp(ctx context.Context, obj *MultishareInstance) (*filev1beta1.Operation, error)
 	StartResizeMultishareInstanceOp(ctx context.Context, obj *MultishareInstance) (*filev1beta1.Operation, error)
+	ListShares(ctx context.Context, filter *ListFilter) ([]*Share, error)
+	GetShare(ctx context.Context, obj *Share) (*Share, error)
 	StartCreateShareOp(ctx context.Context, obj *Share) (*filev1beta1.Operation, error)
 	StartDeleteShareOp(ctx context.Context, obj *Share) (*filev1beta1.Operation, error)
 	StartResizeShareOp(ctx context.Context, obj *Share) (*filev1beta1.Operation, error)
-	WaitForOpWithOpts(ctx context.Context, opts PollOpts) error
+	WaitForOpWithOpts(ctx context.Context, op string, opts PollOpts) error
+	GetOp(ctx context.Context, op string) (*filev1beta1.Operation, error)
+	IsOpDone(op *filev1beta1.Operation) (bool, error)
 }
 
 type gcfsServiceManager struct {
@@ -491,9 +502,21 @@ func (manager *gcfsServiceManager) waitForOp(ctx context.Context, op *filev1beta
 	})
 }
 
+// TODO: unify this function behavior with IsOpDone
 func isOpDone(op *filev1beta1.Operation) (bool, error) {
 	if op == nil {
 		return false, nil
+	}
+	if op.Error != nil {
+		return true, fmt.Errorf("operation %v failed (%v): %v", op.Name, op.Error.Code, op.Error.Message)
+	}
+	return op.Done, nil
+}
+
+func (manager *gcfsServiceManager) IsOpDone(op *filev1beta1.Operation) (bool, error) {
+	// TODO: verify this behavior with filestore
+	if op == nil {
+		return true, nil
 	}
 	if op.Error != nil {
 		return true, fmt.Errorf("operation %v failed (%v): %v", op.Name, op.Error.Code, op.Error.Message)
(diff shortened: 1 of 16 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -487,13 +498,13 @@ func (manager *gcfsServiceManager) waitForOp(ctx context.Context, op *filev1beta
...
-		return isOpDone(pollOp)
+		return manager.IsOpDone(pollOp)
 	})
 }
 
-func isOpDone(op *filev1beta1.Operation) (bool, error) {
+func (manager *gcfsServiceManager) IsOpDone(op *filev1beta1.Operation) (bool, error) {
 	if op == nil {
-		return false, nil
+		return true, nil
```
leiyiz, line 507: in `waitForOp` above, if the `operationsService.Get` call returns nil it'll return false and retry later, now this change will break that behavior iiuc?

pkg/cloud_provider/file/file.go
```diff
@@ -646,6 +658,99 @@ func (manager *gcfsServiceManager) StartResizeShareOp(ctx context.Context, obj *
 	return nil, nil
 }
 
-func (manager *gcfsServiceManager) WaitForOpWithOpts(ctx context.Context, opts PollOpts) error {
+func (manager *gcfsServiceManager) WaitForOpWithOpts(ctx context.Context, op string, opts PollOpts) error {
+	return nil
+}
+
+func (manager *gcfsServiceManager) GetOp(ctx context.Context, op string) (*filev1beta1.Operation, error) {
```
leiyiz, line 665: same pattern in `waitForOp` above, could substitute with this method

pkg/cloud_provider/file/file.go
```diff
@@ -646,6 +658,99 @@ func (manager *gcfsServiceManager) StartResizeShareOp(ctx context.Context, obj *
...
+	}
+
+	mismatches := []string{}
+	if a.Name != b.Name {
+		mismatches = append(mismatches, "tier")
+	}
+	if util.RoundBytesToGb(a.CapacityBytes) != util.RoundBytesToGb(b.CapacityBytes) {
+		mismatches = append(mismatches, "share size")
+	}
+	if len(mismatches) > 0 {
```
leiyiz, line 749: do we not need to compare parents as well

saikat-royc: it's compared in line 752

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -17,27 +17,550 @@ limitations under the License.
...
+		if item != nil {
+			return nil, status.Errorf(codes.Aborted, "Share %q not ready, found op %q", w.share.Name, item.Name)
+		}
+
+		// Start op and update cache
+		op, err := m.cloud.File.StartDeleteShareOp(ctx, w.share)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, err.Error())
+		}
+		m.cache.AddShareOp(instanceSCPrefix, util.CreateShareKey(w.share.Parent.Project, w.share.Parent.Location, w.share.Parent.Name, w.share.Name), util.OpInfo{Name: op.Name, Type: util.ShareExpand})
```
leiyiz, line 346: the type should be `shareDelete` here.
Merge this case with `ShareExpand` and pass in w.OpType in instead

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -17,27 +17,550 @@ limitations under the License.
...
+	if opInfo == nil {
+		return nil, util.StatusUnknown, nil
+	}
+
+	op, err := m.cloud.File.GetOp(ctx, opInfo.Name)
+	if err != nil && !file.IsNotFoundErr(err) {
+		return opInfo, util.StatusUnknown, status.Errorf(codes.Internal, "Failed to get operation %q", opInfo.Name)
+	}
+
+	done, err := m.cloud.File.IsOpDone(op)
```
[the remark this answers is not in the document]
saikat-royc: good question. I dont think so, but need to double check with the filestore api implementation.

pkg/csi_driver/multishare_ops_manager.go
saikat-royc: good catch. Changed to use optype. keeping separate case for readbility and since the expand and delete use different functions StartResizeShareOp() / StartDeleteShareOp().

pkg/util/cache.go
```diff
@@ -156,25 +174,156 @@ func (s *ShareOpsMap) Delete(key ShareKey, opName string) error {
...
+
+func (m *StorageClassInfoCache) AddShareOp(scInfoMapKey string, shareKey ShareKey, opInfo OpInfo) {
+	_, ok := m.ScInfoMap[scInfoMapKey]
+	if !ok {
+		m.ScInfoMap[scInfoMapKey] = StorageClassInfo{
+			InstanceMap:    make(InstanceMap),
+			ShareCreateMap: make(ShareCreateMap),
+			ShareOpsMap:    make(ShareOpsMap),
+		}
+	}
```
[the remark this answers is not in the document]
saikat-royc: agree. Will change it to return error.

pkg/csi_driver/multishare_controller.go
```diff
@@ -65,3 +157,225 @@ func (m *MultishareController) ControllerExpandVolume(ctx context.Context, req *
...
+	}
+
+	if lSet {
+		if rCap == 0 {
+			// request not set
+			return lCap, nil
+		}
+		// request set, round up to min
+		return rCap, nil
+	}
```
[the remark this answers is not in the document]
saikat-royc: Ack. Simplified the check now

pkg/csi_driver/multishare_controller.go
```diff
@@ -65,3 +157,225 @@ func (m *MultishareController) ControllerExpandVolume(ctx context.Context, req *
...
+		Name:          name,
+		Parent:        parent,
+		CapacityBytes: targetSizeBytes,
+		Labels:        extractShareLabels(req.Parameters),
+	}, nil
+}
+
+func (m *MultishareController) pickRegion(top *csi.TopologyRequirement) (string, error) {
+	if top == nil {
+		return m.cloud.Zone, nil
```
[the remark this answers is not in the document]
saikat-royc: ack. fixed it

pkg/cloud_provider/file/file.go
saikat-royc: getop only does GET call for op.

WaitForOpWithOpts will do a get and poll for complete sttaus.

pkg/cloud_provider/file/file.go
```diff
@@ -646,6 +658,99 @@ func (manager *gcfsServiceManager) StartResizeShareOp(ctx context.Context, obj *
 	return nil, nil
 }
 
-func (manager *gcfsServiceManager) WaitForOpWithOpts(ctx context.Context, opts PollOpts) error {
+func (manager *gcfsServiceManager) WaitForOpWithOpts(ctx context.Context, op string, opts PollOpts) error {
```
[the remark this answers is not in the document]
saikat-royc: done

pkg/cloud_provider/file/file.go
saikat-royc: I did think about it. I think if the op is nil, we should return success. This may become more relevant in multishare case, where we may have an op in cache, which has been in the cache for a long time say O(days), and eventually got garbage collected in filestore side. GET would return 404. we need to treat that op as complete and clear cache.

Eventually I plan to unify the behavior and use a single waitforop. For now I didnt want to change the single share instance code path.

I will verify with filestore team, on this behavior.

leiyiz: In order to not change the non-multishare behavior should we create another `IsOpDone` instead of converting the original `isOpDone` which is called by `waitForOp` in line 495

pkg/cloud_provider/file/file.go
leiyiz: Not talking about `WaitForOpWithOpts` but `waitForOp` in line 495.

leiyiz: Not talking about `WaitForOpWithOpts` but `waitForOp` in line 495.

pkg/cloud_provider/file/file.go
leiyiz: line 752 seems to compare `Name` not `Parent`

pkg/cloud_provider/file/file.go
saikat-royc: Ack

saikat-royc: Done

PR #231, opened 2022-03-01
add 404 to user error

amacaskill:
**What type of PR is this?**
/kind bug

This PR distinguishes the user errors from the other internal errors that are returned by CreateSnapshot. Right now, we are being alerted for CreateSnapshot failures that are triggered by a user error:  

`
GRPC error: rpc error: code = Internal desc = Create Backup operation failed: googleapi: Error 404: The resource "projects/itc-iris-staging/locations/europe-west1-b/instances/instace-name" was not found., notFound
`

 This is a user error  that will now have code InvalidArgument instead of Internal. Internal error codes are counted as failures, but InvalidArgument error codes are counted as successes, so this changes user errors to InvalidArgument code. 

```release-note
Users will now see the InvalidArgument error code for the 404 googleapi errors caused by invalid arguments. 
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -494,17 +494,20 @@ func IsNotFoundErr(err error) bool {
 }
 
 // This function returns true if the error is a googleapi error caused by users, such as
-// Error 429: Quota limit exceeded, Error 403: Permission Denied, and Error 400: Bad Request
+// Error 429: Quota limit exceeded, Error 403: Permission Denied, Error 400: Bad Request,
+// Error 404: Not found.
 func IsUserError(err error) bool {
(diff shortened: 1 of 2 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -494,17 +494,20 @@ func IsNotFoundErr(err error) bool {
 }
 
 // This function returns true if the error is a googleapi error caused by users, such as
-// Error 429: Quota limit exceeded, Error 403: Permission Denied, and Error 400: Bad Request
+// Error 429: Quota limit exceeded, Error 403: Permission Denied, Error 400: Bad Request,
+// Error 404: Not found.
 func IsUserError(err error) bool {
```
leiyiz, line 499: consider writing a unit test for this function?

PR #240, opened 2022-03-23
fix inconsistency in definition of paramMultishareInstanceScLabel

leiyiz:
**What type of PR is this?**
 /kind bug

there are conflicting const of `paramMultishareInstanceScLabel` and `util. ParamMultishareInstanceScLabel` having different values, specifically the latter has incorrect definition

Fixes #

```release-note
NONE
```


saikat-royc:
the change is LGTM. ut needs to be fixed.

PR #249, opened 2022-04-08
fix error code mappings

amacaskill:
**What type of PR is this?**
Follow up to #206 which had an invalid http mapping. It checked that the grpc code was in the map, not that the http error code is in the map (which is what the googleapi error has in code.  Also had to remove the error wrapping from the CreateInstance and CreateInstanceFromBackupSource functions so that casting to googleapi error is not nil. 

/kind cleanup

```release-note
returns InvalidArgument instead of Internal error code for some common user errors which is visible to users with a kubectl describe command 
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -29,7 +29,6 @@ import (
 	"github.com/golang/glog"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
-	"google.golang.org/genproto/googleapis/rpc/code"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
@@ -212,17 +211,20 @@ func (manager *gcfsServiceManager) CreateInstance(ctx context.Context, obj *Serv
 		betaObj.Labels)
 	op, err := manager.instancesService.Create(locationURI(obj.Project, obj.Location), betaObj).InstanceId(obj.Name).Context(ctx).Do()
 	if err != nil {
-		return nil, fmt.Errorf("CreateInstance operation failed: %v", err)
+		glog.Errorf("CreateInstance operation failed for instance %s: %v", obj.Name, err)
+		return nil, err
 	}
 
 	glog.V(4).Infof("For instance %v, waiting for create instance op %v to complete", obj.Name, op.Name)
 	err = manager.waitForOp(ctx, op)
 	if err != nil {
-		return nil, fmt.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		glog.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		return nil, err
 	}
 	instance, err := manager.GetInstance(ctx, obj)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get instance after creation: %v", err)
+		glog.Errorf("failed to get instance after creation: %v", err)
+		return nil, err
 	}
 	return instance, nil
 }
@@ -263,17 +265,20 @@ func (manager *gcfsServiceManager) CreateInstanceFromBackupSource(ctx context.Co
 		instance.FileShares[0].SourceBackup)
 	op, err := manager.instancesService.Create(locationURI(obj.Project, obj.Location), instance).InstanceId(obj.Name).Context(ctx).Do()
 	if err != nil {
-		return nil, fmt.Errorf("CreateInstance operation failed: %v", err)
+		glog.Errorf("CreateInstance operation failed for instance %v: %v", obj.Name, err)
+		return nil, err
 	}
 
 	glog.V(4).Infof("For instance %v, waiting for create instance op %v to complete", obj.Name, op.Name)
 	err = manager.waitForOp(ctx, op)
 	if err != nil {
-		return nil, fmt.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		glog.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		return nil, err
 	}
 	serviceInstance, err := manager.GetInstance(ctx, obj)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get instance after creation: %v", err)
+		glog.Errorf("failed to get instance after creation: %v", err)
+		return nil, err
 	}
 	return serviceInstance, nil
 }
@@ -474,7 +479,8 @@ func (manager *gcfsServiceManager) CreateBackup(ctx context.Context, obj *Servic
 	glog.V(4).Infof("Creating backup object %+v for the URI %v", *backupobj, backupUri)
 	opbackup, err := manager.backupService.Create(locationURI(obj.Project, region), backupobj).BackupId(backupName).Context(ctx).Do()
 	if err != nil {
-		return nil, fmt.Errorf("Create Backup operation failed: %v", err)
+		glog.Errorf("Create Backup operation failed: %v", err)
+		return nil, err
 	}
 
 	glog.V(4).Infof("For backup uri %s, waiting for backup op %v to complete", backupUri, opbackup.Name)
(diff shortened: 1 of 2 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -591,13 +590,13 @@ func IsUserError(err error) bool {
 	if !ok {
 		return false
 	}
-	userErrors := map[code.Code]bool{
-		code.Code_RESOURCE_EXHAUSTED: true,
-		code.Code_PERMISSION_DENIED:  true,
-		code.Code_INVALID_ARGUMENT:   true,
-		code.Code_NOT_FOUND:          true,
+	userErrors := map[int]bool{
+		http.StatusTooManyRequests: true,
```
leiyiz, line 594: have you tried this change out manually?

leiyiz: would it make sense to check on both http error code and grpc error code? based on response from Filestore API there's both info [available](https://screenshot.googleplex.com/C5Wue5uNf9JukV8).

amacaskill: I tried it out for everything except filestore api not enabled because it won't let me disable mine on my current project because I have resources still using it. From the error codes I am filtering out, Http error 400 is the only error that has multiple grpc codes, and all the http 400 error codes are user errors so we could do that but it isn't necessary

leiyiz: make sense

amacaskill:
Manually tested error 400 : 
```GRPC error: rpc error: code = InvalidArgument desc = googleapi: Error 400: network 'djakefault' does not exist., badRequest
```
```GRPC error: rpc error: code = InvalidArgument desc = googleapi: Error 400: capacity must be at least 2560 GB., badRequest
```

Wasn't able to test 403 or 429 still trying to test 404 for Create snapshot.

leiyiz:
Yeah that makes sense, if we can't easily reproduce it, it's probably good to remove it as a user error

pkg/cloud_provider/file/file.go
```diff
@@ -474,13 +479,15 @@ func (manager *gcfsServiceManager) CreateBackup(ctx context.Context, obj *Servic
...
-		return nil, fmt.Errorf("Create Backup operation failed: %v", err)
+		glog.Errorf("Create Backup operation failed")
+		return nil, err
 	}
 
 	glog.V(4).Infof("For backup uri %s, waiting for backup op %v to complete", backupUri, opbackup.Name)
 	err = manager.waitForOp(ctx, opbackup)
 	if err != nil {
-		return nil, fmt.Errorf("WaitFor CreateBackup op %s for source instance %v, backup uri: %v, operation failed: %v", opbackup.Name, backupSource, backupUri, err)
+		glog.Errorf("WaitFor CreateBackup op %s for source instance %v, backup uri: %v, operation failed", opbackup.Name, backupSource, backupUri)
```
leiyiz, line 489: What's the reason for logging the errror here instead of adding more info and return it?

pkg/cloud_provider/file/file.go
```diff
@@ -212,17 +211,20 @@ func (manager *gcfsServiceManager) CreateInstance(ctx context.Context, obj *Serv
 		betaObj.Labels)
 	op, err := manager.instancesService.Create(locationURI(obj.Project, obj.Location), betaObj).InstanceId(obj.Name).Context(ctx).Do()
 	if err != nil {
-		return nil, fmt.Errorf("CreateInstance operation failed: %v", err)
+		glog.Errorf("CreateInstance operation failed")
```
saikat-royc, line 214: maybe put the error in the log too

pkg/cloud_provider/file/file.go
```diff
@@ -212,17 +211,20 @@ func (manager *gcfsServiceManager) CreateInstance(ctx context.Context, obj *Serv
...
+		return nil, err
 	}
 
 	glog.V(4).Infof("For instance %v, waiting for create instance op %v to complete", obj.Name, op.Name)
 	err = manager.waitForOp(ctx, op)
 	if err != nil {
-		return nil, fmt.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		glog.Errorf("WaitFor CreateInstance op %s failed", op.Name)
```
saikat-royc, line 221: same here and other places

pkg/csi_driver/controller.go
```diff
@@ -736,9 +736,7 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
 	backupObj, err := s.config.fileService.CreateBackup(ctx, filer, req.Name, util.GetBackupLocation(req.GetParameters()))
 	if err != nil {
 		glog.Errorf("Create snapshot for volume Id %s failed: %v", volumeID, err)
-		if file.IsUserError(err) {
-			return nil, status.Error(codes.InvalidArgument, err.Error())
-		} else {
+		if err != nil {
```
saikat-royc, line 739: not needed?

pkg/cloud_provider/file/file.go
saikat-royc: you can delete any existing filestore resources in the project and try out `gcloud services disable  "file.googleapis.com" --force`

saikat-royc: Also thinking a bit I think 429 should be treated as user error. Let's discuss this a bit more offline. We ideally want to be alerted on quota exceed errors.

pkg/csi_driver/controller.go
amacaskill: Did you see [comment](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/249#issuecomment-1095600739). I wasn't able to reproduce this error. When I tried I got `Failed to create snapshot content with error failed to get input parameters to create snapshot backup-source-pvc: "the PVC source-pvc is not yet bound to a PV, will not attempt to take a snapshot"`

pkg/cloud_provider/file/file.go
amacaskill: I tried force as well and it still won't let me 
`ERROR: (gcloud.services.disable) The operation "operations/acf.p17-876590250267-9939fe22-4d40-49ec-83c4-fc57675d9e58" resulted in a failure "[Error in service 'file.googleapis.com': Not ready to deactivate the service on this project; ensure there are no more resources managed by this service.`

pkg/cloud_provider/file/file.go
```diff
@@ -212,17 +211,20 @@ func (manager *gcfsServiceManager) CreateInstance(ctx context.Context, obj *Serv
 		betaObj.Labels)
 	op, err := manager.instancesService.Create(locationURI(obj.Project, obj.Location), betaObj).InstanceId(obj.Name).Context(ctx).Do()
 	if err != nil {
-		return nil, fmt.Errorf("CreateInstance operation failed: %v", err)
+		glog.Errorf("CreateInstance operation failed: %v", err)
```
saikat-royc, line 214: capture the instance details here

pkg/cloud_provider/file/file.go
```diff
@@ -263,17 +265,20 @@ func (manager *gcfsServiceManager) CreateInstanceFromBackupSource(ctx context.Co
...
+		return nil, err
 	}
 
 	glog.V(4).Infof("For instance %v, waiting for create instance op %v to complete", obj.Name, op.Name)
 	err = manager.waitForOp(ctx, op)
 	if err != nil {
-		return nil, fmt.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
+		glog.Errorf("WaitFor CreateInstance op %s failed: %v", op.Name, err)
```
saikat-royc, line 275: capture the instance details here

pkg/csi_driver/controller.go
```diff
@@ -736,9 +738,7 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
 	backupObj, err := s.config.fileService.CreateBackup(ctx, filer, req.Name, util.GetBackupLocation(req.GetParameters()))
 	if err != nil {
```
saikat-royc, line 739: remove

amacaskill:
also  manually tested 403 error: 
```
Warning  ProvisioningFailed    12s (x5 over 27s)  filestore.csi.storage.gke.io_gke-7f715c2c9f5d4c938326-7586-15c5-vm_29746fe9-8b54-4a5a-81f0-0e650c2dd156  failed to provision volume with StorageClass "filestore-example": rpc error: code = InvalidArgument desc = googleapi: Error 403: Cloud Filestore API has not been used in project 876590250267 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/file.googleapis.com/overview?project=876590250267 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```

PR #268, opened 2022-05-03
Implement e2e test for NodeGetVolumeStats

tyuchn:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Add test for NodeGetVolumeStats in resize e2e test.
```

### DIFF
--- a/test/e2e/tests/resize_e2e_test.go
+++ b/test/e2e/tests/resize_e2e_test.go
@@ -16,8 +16,10 @@ package tests
 
 import (
 	"fmt"
+	"math/rand"
 	"path/filepath"
 	"strings"
+	"time"
 
 	filev1beta1 "google.golang.org/api/file/v1beta1"
 	"k8s.io/apimachinery/pkg/util/uuid"
@@ -39,6 +41,7 @@ const (
 	defaultTier          = "STANDARD"
 	defaultNetwork       = "default"
 	minVolumeSize  int64 = 1 * util.Tb
+	defaultEpsilon int64 = 70 * util.Gb // when a 1T instance was created, the actual size is 1007G, and only 956G is available to non-root users, so there will be ~70 Gb of space unavailable
 )
 
 var _ = Describe("Google Cloud Filestore CSI Driver", func() {
@@ -51,7 +54,7 @@ var _ = Describe("Google Cloud Filestore CSI Driver", func() {
 
 		validateDisk(diskInfo)
 
-		writeAndReadDisk(diskInfo)
+		writeAndReadDisk(diskInfo, testContext)
 
 		offlineResizeDisk(diskInfo)
 
@@ -59,7 +62,7 @@ var _ = Describe("Google Cloud Filestore CSI Driver", func() {
 	})
 })
 
-func writeAndReadDisk(di *DiskInfo) {
+func writeAndReadDisk(di *DiskInfo, tc *remote.TestContext) {
 	var err error
 	instance := di.TestCtx.Instance
 	volName := di.Name
@@ -97,6 +100,8 @@ func writeAndReadDisk(di *DiskInfo) {
 	}()
 
 	validateRead(secondPublishDir, testFileName, "test", instance)
+
+	validateVolumeStats(publishDir, di.Volume.VolumeId, instance, tc)
 }
 
 func onlineResizeDisk(di *DiskInfo) {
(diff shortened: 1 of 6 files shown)


test/e2e/tests/resize_e2e_test.go
```diff
@@ -39,6 +39,7 @@ const (
 	defaultTier          = "STANDARD"
 	defaultNetwork       = "default"
 	minVolumeSize  int64 = 1 * util.Tb
+	defaultEpsilon int64 = 70 * util.Gb
```
saikat-royc, line 42: can you put a comment explaining some background reasoning of this number

test/e2e/tests/resize_e2e_test.go
```diff
@@ -288,6 +290,17 @@ func validateRead(publishDir, testFileName, testFileContents string, instance *r
 	Expect(strings.TrimSpace(string(readContents))).To(Equal(testFileContents))
 }
 
+func validateVolumeStats(publishDir, volID string, tc *remote.TestContext) {
+	available, capacity, used, inodesFree, inodes, inodesUsed, err := tc.Client.NodeGetVolumeStats(volID, publishDir)
+	Expect(err).To(BeNil(), "Failed to get node volume stats: %v", err)
+	Expect(equalWithinEpsilon(available, minVolumeSize, defaultEpsilon)).To(BeTrue())
+	Expect(equalWithinEpsilon(capacity, minVolumeSize, defaultEpsilon)).To(BeTrue())
+	Expect(equalWithinEpsilon(used, 0, defaultEpsilon)).To(BeTrue())
```
leiyiz, line 298: there have been file written to the volume, so here used should not be 0

test/e2e/tests/resize_e2e_test.go
tyuchn: Done

test/e2e/tests/resize_e2e_test.go
tyuchn: Yes, but it's hard to quantitive the usage. So let's set a threshold of 70G at this point, and modify it latter if things changed.

leiyiz: the usage should be deterministic in this case right? it's just "test" with whatever overhead a file has. There's no reason to have 70G of epsilon here and the usage should not be 0 either.

test/e2e/tests/resize_e2e_test.go
```diff
@@ -41,7 +41,7 @@ const (
 	defaultTier          = "STANDARD"
 	defaultNetwork       = "default"
 	minVolumeSize  int64 = 1 * util.Tb
-	defaultEpsilon int64 = 70 * util.Gb // when a 1T instance was created, there will be 68-70 Gb of space unavailable
+	defaultEpsilon int64 = 70 * util.Gb // when a 1T instance was created, the actual size is 1007G, and only 956G is available to non-root users, so there will be ~70 Gb of space unavailable
```
leiyiz, line 44: Should we instead count the actual size instead of space available to non-root users? Can a pod be root user as well? @saikat-royc

test/e2e/tests/resize_e2e_test.go
```diff
@@ -290,12 +293,18 @@ func validateRead(publishDir, testFileName, testFileContents string, instance *r
...
+	fileSize := 1 + rand.Int63n(100)
+	err := testutils.GenerateRandomFile(instance, publishDir, fileSize)
+	Expect(err).To(BeNil(), "Failed to generate file")
+
 	available, capacity, used, inodesFree, inodes, inodesUsed, err := tc.Client.NodeGetVolumeStats(volID, publishDir)
 	Expect(err).To(BeNil(), "Failed to get node volume stats: %v", err)
 	Expect(equalWithinEpsilon(available, minVolumeSize, defaultEpsilon)).To(BeTrue())
 	Expect(equalWithinEpsilon(capacity, minVolumeSize, defaultEpsilon)).To(BeTrue())
-	Expect(equalWithinEpsilon(used, 0, defaultEpsilon)).To(BeTrue())
+	Expect(used).To(Equal(util.MbToBytes(fileSize)))
```
leiyiz, line 307: have you ran this to verify that the `used` be exactly the same with `fileSize` to the byte? do we need to add a small epsilon here of like 0.5%?

tyuchn: Yes, it's verified. As long as the file size is a multiple of 1Mb, `NodeGetVolumeStats` will return the exact number we want.

test/e2e/tests/resize_e2e_test.go
saikat-royc: as discussed let's report all

test/e2e/tests/resize_e2e_test.go
```diff
@@ -39,6 +41,7 @@ const (
 	defaultTier          = "STANDARD"
 	defaultNetwork       = "default"
 	minVolumeSize  int64 = 1 * util.Tb
+	defaultEpsilon int64 = 70 * util.Gb // when a 1T instance was created, the actual size is 1007G, and only 956G is available to non-root users, so there will be ~70 Gb of space unavailable
```
leiyiz, line 44: since we changed how the free space is reported, we can change this epsilon to 20G

tyuchn: Fixed

PR #269, opened 2022-05-06
align instance resize target bytes

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Instance resize expects the target size to be a multiple of the instance step size.

Fixes #

more instance resize UTs will be added once I am able to test the change on a real setup

```release-note
align instance resize target bytes
```

### DIFF
--- a/pkg/csi_driver/multishare_ops_manager.go
+++ b/pkg/csi_driver/multishare_ops_manager.go
@@ -336,11 +336,13 @@ func (m *MultishareOpsManager) instanceNeedsExpand(ctx context.Context, share *f
 	for _, s := range shares {
 		sumShareBytes = sumShareBytes + s.CapacityBytes
 	}
-	// TODO: Check if we need to align the increment to step size.
+
 	var remainingBytes int64
 	remainingBytes = share.Parent.CapacityBytes - sumShareBytes
 	if remainingBytes < share.CapacityBytes {
-		return true, (share.CapacityBytes + sumShareBytes), nil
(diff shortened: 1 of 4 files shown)


pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -418,7 +420,13 @@ func (m *MultishareOpsManager) checkAndStartInstanceDeleteOrShrinkWorkflow(ctx c
 		totalShareCap += share.CapacityBytes
 	}
 	if totalShareCap < instance.CapacityBytes && instance.CapacityBytes > util.MinMultishareInstanceSizeBytes {
-		instance.CapacityBytes = util.Max(totalShareCap, util.MinMultishareInstanceSizeBytes)
+		targetShrinkSizeBytes := util.AlignBytes(totalShareCap, util.GbToBytes(instance.CapacityStepSizeGb))
+		targetShrinkSizeBytes = util.Max(targetShrinkSizeBytes, util.MinMultishareInstanceSizeBytes)
```
leiyiz, line 424: why do we need this check? isn't the max size and min size all aligned with the step size?

saikat-royc: say sum of shares size is 500G, and step size is 256G, we want the targetShrinkSizeBytes to be 1TiB (and not 512 GiB)

PR #271, opened 2022-05-18
first list instances before list share

leiyiz:
**What type of PR is this?**
 /kind bug

Filestore side listShare across regions and instances doesn't work yet, work around it by list all instances if the `instanceName` filter is `-`

Fixes #

```release-note
NONE
```


pkg/cloud_provider/file/file.go
```diff
@@ -860,39 +860,55 @@ func (manager *gcfsServiceManager) GetShare(ctx context.Context, obj *Share) (*S
...
+
+				s := &Share{
+					Name: shareName,
+					Parent: &MultishareInstance{
+						Name:     instanceName,
+						Project:  project,
+						Location: location,
+					},
+					MountPointName: sobj.MountName,
+					CapacityBytes:  sobj.CapacityGb * util.Gb,
```
saikat-royc, line 905: can you also capture labels and state in this object.
Also similar thing to be done in GetShare() https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/cloud_provider/file/file.go#L853

leiyiz: done

PR #275, opened 2022-05-20
support ip reservation for multi-share filestore

leiyiz:
**What type of PR is this?**
/kind bug
/kind feature

- multishare Filestore CSI does not support FIlestore Ip reservation.
- current Filestore CSI ip reservation will not work if there's multi-share instance present in the driver
- fixed a bug where if the endpoint is not specified, the host URL will not correctly point to prod endpoint

Fixes #

```release-note
support ip reservation for multi-share filestore
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -166,6 +166,7 @@ var (
 
 func NewGCFSService(version string, client *http.Client, endpoint string) (Service, error) {
 	ctx := context.Background()
+
 	fileService, err := filev1beta1.NewService(ctx, option.WithHTTPClient(client))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -166,14 +166,16 @@ var (
 
 func NewGCFSService(version string, client *http.Client, endpoint string) (Service, error) {
 	ctx := context.Background()
-	fileService, err := filev1beta1.NewService(ctx, option.WithHTTPClient(client))
+	basepath := createFilestoreEndpointUrlBasePath(endpoint)
+	glog.Infof("Using endpoint %q", basepath)
+
+	fileService, err := filev1beta1.NewService(ctx, basepath, option.WithHTTPClient(client))
```
saikat-royc, line 172: for single share we do not want to tweak the env based on the flag.

pkg/csi_driver/multishare_controller.go
```diff
@@ -87,6 +88,33 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
...
+			Project:  instance.Project,
+			Name:     instance.Name,
+			Location: instance.Location,
+			Tier:     instance.Tier,
+		}, reservedIPV4CIDR)
+
+		// Possible cases are 1) CreateInstanceAborted, 2)CreateInstance running in background
+		// The ListInstances response will contain the reservedIPRange if the operation was started
+		// In case of abort, the CIDR IP is released and available for reservation
+		defer m.controllerServer.config.ipAllocator.ReleaseIPRange(reservedIPRange)
```
saikat-royc, line 109: note this defer is executed at the end of the function call. this means lock will be released only after instance + share create complete in worst case.  we can technically release lock at the end of instance create. However since share create is relatively much faster, this should be fine for now. something to keep in mind.

PR #311, opened 2022-06-21
Check for instance READY state before placing a share on it.

tyuchn:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/pkg/csi_driver/multishare_ops_manager.go
+++ b/pkg/csi_driver/multishare_ops_manager.go
@@ -324,6 +324,11 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, ins
 	var readyEligibleInstances []*file.MultishareInstance
 	nonReadyInstanceCount := 0
 	for _, instance := range instances {
+		if instance.State != "READY" {
+			klog.V(5).Infof("Instance %s/%s/%s with state %s is not eligible", instance.Project, instance.Location, instance.Name, instance.State)
+			continue
+		}
+
 		op, err := containsOpWithInstanceTargetPrefix(instance, ops)
 		if err != nil {
 			klog.Errorf("failed to check eligibility of instance %s", instance.Name)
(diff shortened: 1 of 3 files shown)


pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -324,6 +324,11 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, ins
 	var readyEligibleInstances []*file.MultishareInstance
 	nonReadyInstanceCount := 0
 	for _, instance := range instances {
+		if instance.State != "READY" {
```
saikat-royc, line 327: we should consider CREATING as a non ready instance too

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -345,16 +359,6 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, ins
...
-		if op.Type == util.InstanceDelete {
-			continue
-		}
-
-		if instance.State == "DELETING" || instance.State == "ERROR" {
-			klog.V(5).Infof("Instance %s/%s/%s with state %s is not eligible", instance.Project, instance.Location, instance.Name, instance.State)
-			continue
-		}
-
 		klog.Infof("Instance %s/%s/%s is not ready with ongoing operation %s type %s", instance.Project, instance.Location, instance.Name, op.Id, op.Type.String())
```
saikat-royc, line 362: let's also log the state here

tyuchn: Done

pkg/csi_driver/multishare_ops_manager.go
jingxu97: what if the instance is in the process of restarting? do we need any logic to handle this case specifically or "not ready" can work for this case too?

PR #321, opened 2022-06-29
register process start time metric for core filestore driver container

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Filestore CSI driver emits multishare operations latency histogram [metric](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/v1.2.7/pkg/metrics/metrics.go#L50) (which is cummulative). For the metric to work properly, the process start time metric needs to be registered correctly. The csi sidecars do a similar thing for the csi_sidecar_operation_seconds metric. See conversation in b/168620097#comment32. The snippet is 

```
Every application starts growing all cumulative metrics from 0. When application crashes, counter resets. That is true for prometheus metrics. For correctly handling such situations Monarch expects to get information about start time of an application (point of time when counter was reset). If not provided prometheus-to-sd will leave this field empty which means reset time will be equal to 1970. In such case cumulative metrics will be processed incorrectly.
```

Additional note:
Related [PR](https://github.com/kubernetes/kubernetes/pull/98391) to fix the RegisterProcessStartTime which is used by [csi-lib-utils](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/metrics/metrics.go#L224), the same lib used by sidecars. The filestorecsi is already using [0.24.1](https://github.com/kubernetes/component-base/blob/v0.24.1/metrics/processstarttime.go#L37) of k8s.io/component-base ([go.mod](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/v1.2.7/go.mod#L33)), so only a single call to RegisterProcessStartTime() is sufficient, as it should successfully register metric and also record the start time.

Also verified it manually, that the controller emits the process_start_time metric
```
saikatroyc@gke-cluster-5-default-pool-839e3bc1-sz75 ~ $ curl localhost:22024/metrics
# HELP process_start_time_seconds [ALPHA] Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.65648302501e+09
```

Fixes #

```release-note
register process start time metric for core filestore driver container
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -65,6 +65,7 @@ func NewMetricsManager() *MetricsManager {
 	mm := &MetricsManager{
 		registry: metrics.NewKubeRegistry(),
 	}
+	metrics.RegisterProcessStartTime(mm.registry.Register)
 	mm.registry.MustRegister(operationSeconds)
 	return mm
 }
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -1,0 +1,26 @@
+package metrics
+
+import (
+	"testing"
+)
+
+const (
+	ProcessStartTimeMetric = "process_start_time_seconds"
+)
+
+func TestProcessStartTimeMetricExist(t *testing.T) {
+	mm := NewMetricsManager()
+	metricsFamilies, err := mm.GetRegistry().Gather()
+	if err != nil {
+		t.Fatalf("Error fetching metrics: %v", err)
+	}
+
+	// check 'process_start_time_seconds' metric exist
+	for _, metricsFamily := range metricsFamilies {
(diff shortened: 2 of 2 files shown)


leiyiz (commented):
/LGTM

tyuchn:
May I ask where port number `22024` was defined? Looks like we didn't specify HttpPoint for OSS driver, is it intended? @saikat-royc 
https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/f5a6b40ae91f77940c4a4ec61df33be9a239b2e6/deploy/kubernetes/base/controller/controller.yaml#L106-L112

PR #322, opened 2022-06-30
Check for GiB aligned sizes for multishare volumes

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Without this PR, the driver fails in a lazy fashion after performing API calls. 
For exapand, the filestore update API is called with a capacity aligned to lower Gib, the op is a no-op from filestore side, and then CSI driver fails as it does not find expected size.

For create, the filestore op completes with the lower GiB number. and then external provisioner deletes the PV and retries. the loop continues forever

Fixes #

```release-note

```


pkg/csi_driver/multishare_controller.go
```diff
@@ -80,10 +80,13 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	_, err = getShareRequestCapacity(req.GetCapacityRange())
+	reqBytes, err := getShareRequestCapacity(req.GetCapacityRange())
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	if !util.IsAligned(reqBytes, util.Gb) {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Invalid requested size(bytes) %d", reqBytes))
```
leiyiz, line 88: consider add more info here: "size need to be multiply of 1GiB"

saikat-royc: done

leiyiz (commented):
/LGTM

PR #325, opened 2022-07-15
Update image buster for multiarch (arm) support

mattcary:
/kind bug

The old launcher.gcr.io image is not multiarch. The PR updates the base image for the driver to debian buster, which is multiarch.

This also adds the netbase apt package which is now necessary for nfs-common (it apparently was not needed or was already installed in the launcher.gcr.io image).

```release-note
Support for ARM nodes.
```

/assign @saikat-royc


mattcary:
Provisioning is failling with the same authentication error I saw trying to run this locally:

```
textPayload: "error fetching initial token: oauth2: cannot fetch token: Post "https://oauth2.googleapis.com/token": x509: certificate signed by unknown authority"
```

leiyiz (commented):
/lgtm

PR #340, opened 2022-07-29
consume official v1beta1 go client in preparation for GA

leiyiz:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```


vendor/google.golang.org/api/file/v1beta1/file-gen.go
```diff
@@ -86,7 +87,30 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+	if endpoint != "" {
+		s.BasePath = endpoint
+	}
+	return s, nil
+}
+
+// NewServiceBasePath creates a new Service.
```
saikat-royc, line 111:
```
	svc, err := v1beta1cloudfiler.NewService(ctx, option.WithHTTPClient(httpClient))
	if err != nil {
		return err
	}
	if filestoreEndpoint == AutopushEndpoint {
		svc.BasePath = AutopushBasePath
	} else if filestoreEndpoint == StagingEndpoint {
		svc.BasePath = StagingBasePath
	} else {
		svc.BasePath = ProdBasePath
	}
```
seems to work

leiyiz: done

PR #341, opened 2022-07-29
Multishare ip reservation bug fix

tyuchn:
**What type of PR is this?**
>
> /kind api-change
kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When creating multishare instances with ip reservation (suppose the `reserved-ipv4-cidr` is `10.0.0.0/24` in StorageClass), the workflow is:
1. Driver [prepare a dummy instance object](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/54ddbc8640c003c9ce86b5575db51851a8014718/pkg/csi_driver/multishare_controller.go#L96) based on the CSI request.
2. Driver [list all existing single/multishare instances](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/54ddbc8640c003c9ce86b5575db51851a8014718/pkg/csi_driver/multishare_controller.go#L109), get their ip ranges, then calculate an available ip range for the dummy instance object.
3. Driver [list all existing multishare instances again](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/54ddbc8640c003c9ce86b5575db51851a8014718/pkg/csi_driver/multishare_controller.go#L128), check if there's any eligible instance to match the dummy instance (only match for SC prefix so far)
4. If no eligible instance found, the drive will go ahead and create a new filestore instance using the dummy instance object.

At the first iteration, the driver ended up at step#4, and returned an error since instance creation op exceeded the context timeout. 

Then at the second iteration, the driver was stuck at step#2, since an instance with ip range `10.0.0.0/24` is being created. As a result, the driver gets into an endless loop and keeps returning `GRPC error: all of the /24 IP ranges in the cidr 10.0.0.0/24 are reserved`.

We want to avoid such issue, so we need to check but only check for available ip range before creating a new filestore instance.

Fixes #[b/240703744](b/240703744)

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/multishare_ops_manager.go
+++ b/pkg/csi_driver/multishare_ops_manager.go
@@ -50,8 +50,9 @@ type Workflow struct {
 
 // MultishareOpsManager manages the lifecycle of all instance and share operations.
 type MultishareOpsManager struct {
-	sync.Mutex // Lock to perform thread safe multishare operations.
-	cloud      *cloud.Cloud
+	sync.Mutex       // Lock to perform thread safe multishare operations.
+	cloud            *cloud.Cloud
+	controllerServer *controllerServer
 }
 
 func NewMultishareOpsManager(cloud *cloud.Cloud) *MultishareOpsManager {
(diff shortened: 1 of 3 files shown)


pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -52,6 +52,7 @@ type Workflow struct {
 type MultishareOpsManager struct {
 	sync.Mutex // Lock to perform thread safe multishare operations.
 	cloud      *cloud.Cloud
+	controllerServer *controllerServer
```
leiyiz, line 55: looks like format issue here

leiyiz (commented):
/lgtm
/approve

leiyiz:
please rebase your changes :)

PR #347, opened 2022-08-04
Add reserved ip range check and fail earlier if invalid

tyuchn:
**What type of PR is this?**
>
> /kind api-change
kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Regarding [b/240141161](b/240141161), if user specified reserved-ip-range as a direct value, they will get `when using connect mode PRIVATE_SERVICE_ACCESS, if reserved IP range is specified, it must be a named address range instead of direct CIDR value`.  
Right now, there is no check and the error actually occurs from the CreateInstance call which fails when the operation times out because of the user error. We should return an error as soon as the user made a user error, and return an invalid argument.

Fixes #[b/240141161](b/240141161)

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/utils.go
+++ b/pkg/csi_driver/utils.go
@@ -71,3 +71,9 @@ func IsIpWithinRange(ipAddress, ipRange string) (bool, error) {
 	}
 	return ipnet.Contains(net.ParseIP(ipAddress)), nil
 }
+
+// IsCIDR verifies if the given ip range is a valid CIDR value.
+func IsCIDR(ipRange string) bool {
+	_, _, err := net.ParseCIDR(ipRange)
+	return err == nil
+}
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -211,6 +211,9 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		// If the param was not provided, we default reservedIPRange to "" and cloud provider takes care of the allocation
 		if newFiler.Network.ConnectMode == privateServiceAccess {
 			if reservedIPRange, ok := param[paramReservedIPRange]; ok {
+				if IsCIDR(reservedIPRange) {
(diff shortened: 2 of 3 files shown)


leiyiz (commented):
/ok-to-test

pkg/csi_driver/utils.go
```diff
@@ -59,3 +62,13 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 	}
 	return resp, err
 }
+
+// IsValidReservedIpRange verifies if reserved IP range is specified,
+// it must be a named address range instead of direct CIDR value.
+func IsValidReservedIpRange(ipRange string) bool {
```
leiyiz, line 68: maybe this should be named "IsNotCIDR"

pkg/csi_driver/utils.go
```diff
@@ -59,3 +62,13 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
...
+
+// IsNotCIDR verifies if reserved IP range is specified,
+// it must be a named address range instead of direct CIDR value.
+func IsNotCIDR(ipRange string) bool {
+	_, _, err := net.ParseCIDR(ipRange)
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "invalid CIDR address")
+}
```
leiyiz, line 74: actually here how this function is called is probably not concise - here the function reverses the "is something CIDR" and then when it was used the result is !-ed again.

So my suggestion is actually just to write a "IsCIDR" and make the whole thing simpler.

In fact, I believe you can directly treat `if _, _, err := net.ParseCIDR(ipRange); err == nil` as the condition that a string `ipRange` is a CIDR range and false otherwise without needing to match "invalid CIDR address" in the error it returned.

tyuchn: Agree, revised function to IsCIDR.

leiyiz (commented):
/lgtm
/approve

PR #365, opened 2022-08-23
Replace invalid characters in label values

bertinatto:
**What type of PR is this?**
> /kind bug

The backend report multiple errors like this when the CSI Driver is
tested with the External Test Suite:

```
googleapi: Error 400: resource labels are invalid: value "filestore.csi.storage.gke.io66m92" contains invalid character '.' at index 9 for key "kubernetes_io_created-for_pvc_name", badRequest
```

This patch proposes to replace invalid characters, as described here [a], in label values.

[1] https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements

Fixes #

```release-note

```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/controller.go
```diff
@@ -680,10 +681,24 @@ func getZoneFromSegment(seg map[string]string) (string, error) {
...
+		return "", fmt.Errorf("failed to compile regex to match valid labels: %w", err)
+	}
+	return reg.ReplaceAllString(s, "_"), nil
+}
+
 func extractLabels(parameters map[string]string, driverName string) (map[string]string, error) {
 	labels := make(map[string]string)
 	scLables := make(map[string]string)
 	for k, v := range parameters {
+		v, err := replaceInvalidChars(v)
```
saikat-royc, line 698: Ideally the driver should not take such decision to make intrinsic conversion of user provided labels.
We only make such conversion for the driver name since its not a user provided label.

I do see there is some conflict:
k8s naming [rules](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) allow '.', while labels [rules](https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements) do not.

Can you point to which test causes this error?

PR #370, opened 2022-08-25
Replace global listInstance call with regional listInstance calls only for allowed regions

tyuchn:
**What type of PR is this?**
>
> /kind api-change
kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

1. ListMultishareInstance call will return a partial result if any of the regions is unreachable. This will result the csi driver in creating redundant instances. After this fix, ListMultishareInstance call will fail earlier if the specified region is unreachable, and retry at the next csi request.
2. Format revision: replace hard coded strings with consts in unit tests.

Fixes #b/243699081

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/multishare_controller_test.go
+++ b/pkg/csi_driver/multishare_controller_test.go
@@ -467,7 +467,7 @@ func TestGenerateNewMultishareInstance(t *testing.T) {
 				t.Error("expected error, got none")
 			}
 			if !tc.expectErr && err != nil {
-				t.Error("unexpected error")
+				t.Errorf("unexpected error: %q", err)
 			}
 			if !reflect.DeepEqual(filer, tc.expectedInstance) {
 				t.Errorf("got filer %+v, want %+v", filer, tc.expectedInstance)
--- a/pkg/csi_driver/multishare_ops_manager_test.go
+++ b/pkg/csi_driver/multishare_ops_manager_test.go
@@ -30,13 +30,15 @@ import (
 )
 
 const (
-	testInstanceScPrefix = "testinstancescprefix"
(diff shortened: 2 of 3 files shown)


pkg/csi_driver/multishare_controller_test.go
```diff
@@ -467,7 +467,7 @@ func TestGenerateNewMultishareInstance(t *testing.T) {
 				t.Error("expected error, got none")
 			}
 			if !tc.expectErr && err != nil {
-				t.Error("unexpected error")
+				t.Errorf("unexpected error: %q", err)
```
leiyiz, line 470: what is the purpose of this change?

pkg/csi_driver/multishare_ops_manager_test.go
```diff
@@ -30,13 +30,15 @@ import (
...
-	testInstanceScPrefix = "testinstancescprefix"
+	testInstanceScPrefix = "test-prefix"
 	testInstanceName     = "testInstanceName"
 	testShareName        = "testShareName"
+	testVPCNetwork       = "testSharedNetwork"
 )
 
 var (
 	testInstanceHandle = fmt.Sprintf("%s/%s/%s", testProject, testRegion, testInstanceName)
+	testRegions        = []string{"us-central1"}
```
leiyiz, line 41: why hard code region here but use `testRegion` everywhere else

tyuchn: Fixed. It should be replaced with `testRegion`.

leiyiz (commented):
/lgtm
/approve

PR #372, opened 2022-09-02
fix sc across cluster can point to the same instance pool

leiyiz:
**What type of PR is this?**
/kind bug

Fixes #

```release-note
if multishare is enabled, the container now requires "--gke-cluster-name" flag to be set
```


pkg/csi_driver/controller.go
```diff
@@ -76,11 +76,13 @@ const (
 	// User provided labels
 	ParameterKeyLabels = "labels"
 
-	// Keys for tags to attach to the provisioned disk.
+	// Keys for tags to attach to the provisioned Filestore shares and instances.
 	tagKeyCreatedForClaimNamespace = "kubernetes_io_created-for_pvc_namespace"
 	tagKeyCreatedForClaimName      = "kubernetes_io_created-for_pvc_name"
 	tagKeyCreatedForVolumeName     = "kubernetes_io_created-for_pv_name"
 	tagKeyCreatedBy                = "storage_gke_io_created-by"
+	tagKeyClusterName              = "gke_cluster_name"
```
saikat-royc, line 84: maybe storage_gke_io_cluster_name and storage_gke_io_cluster_location ?

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -690,15 +690,14 @@ func (m *MultishareOpsManager) listMatchedInstances(ctx context.Context, req *cs
 // 7, Both source and target instance should have the same connect mode.
 // 8. Both source and target instance should have the same KmsKeyName.
 func isMatchedInstance(source, target *file.MultishareInstance, req *csi.CreateVolumeRequest) (bool, error) {
```
saikat-royc, line 692: please update the comment above the function "Both source and target instance..."

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -690,15 +690,14 @@ func (m *MultishareOpsManager) listMatchedInstances(ctx context.Context, req *cs
...
-	if scPrefixSource, ok := source.Labels[util.ParamMultishareInstanceScLabelKey]; ok {
-		if _, ok := target.Labels[util.ParamMultishareInstanceScLabelKey]; !ok {
-			return false, fmt.Errorf("label %q missing in target instance %+v", util.ParamMultishareInstanceScLabelKey, target)
+	matchLabels := [3]string{util.ParamMultishareInstanceScLabelKey, tagKeyClusterLocation, tagKeyClusterName}
+	for _, labelKey := range matchLabels {
+		if _, ok := target.Labels[labelKey]; !ok {
```
saikat-royc, line 695: what about source missing the key? I think we should handle that too

saikat-royc:
Before submitting this PR we should also verify GKE can actually plumb the cluster name and location to the yamls

saikat-royc:
Please update the description of the PR to provide a background context why these extra tags are needed (also why two separate tags needed)

pkg/csi_driver/multishare_ops_manager_test.go
```diff
@@ -543,14 +553,28 @@ func TestListMatchedInstances(t *testing.T) {
...
+						tagKeyClusterName:                      testClusterName,
+					},
+				},
+				{
+					Name:     "test-instance-3",
+					Project:  "test-project",
+					Location: "us-central1",
+					Labels: map[string]string{
+						util.ParamMultishareInstanceScLabelKey: "testprefix-3",
+						tagKeyClusterLocation:                  testLocation,
```
saikat-royc, line 576: Let's have test cases for mismatch location and mismatched cluster name as well. (If I have not missed it)

pkg/csi_driver/multishare_ops_manager.go
leiyiz: yeah that's exactly what's done here. See L698

pkg/csi_driver/multishare_ops_manager_test.go
leiyiz: done

pkg/csi_driver/multishare_ops_manager.go
saikat-royc: Ack

pkg/csi_driver/controller.go
leiyiz: the reason I didn't use `storage_gke_io_` prefix is because that might be too similar with the `storage_gke_io_created-by` and since we expect customers to change the cluster name and location tag I think simpler, easier to read tag keys make more sense

pkg/csi_driver/multishare_ops_manager.go
leiyiz: done.

pkg/csi_driver/controller.go
saikat-royc: I think the primary indicator of the tag prefix is storage_gke_io_* are used by CSI driver to manage stuff. User are free to modify labels anyway. @mattcary  what are your thoughts on the tag naming?

mattcary: We should be consistent with the pd driver, which uses [storage.gke.io/created-by](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/common/parameters.go#L48) if I read the code correctly.

That's the principle of least surprise at least.

leiyiz: I think the `tagCreatedBy` is inconsistent because [filestore tag key constraint](https://cloud.google.com/filestore/docs/managing-labels#requirements):
> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes

as for the cluster name and location I think having multiple tag key start with `storage_gke_io_ might be confusing: https://screenshot.googleplex.com/Auh94HCt9cHnTFR

mattcary: I think we have to have a google-specific prefix, though, or we're going to clash with customer-provided tags. So no matter we do I think the current UI isn't going to be great and we may need to fix that.

We have an open issue to start putting the PD tags into labels as well as the description, so the encoding issue is going to come up again (labels didn't exist I guess when the csi driver was first made).

I think it's important to have uniformity across our products. And in that context the names always end up a bit clunky.

saikat-royc: Agree. UI should not drive the label names, if should be the other way around. For uniformity, and clear identification of GKE managed tags, let's proceed with storage_gke_io_ prefixed label for the cluster name. storage_gke_io_cluster_name for example

leiyiz: sgtm done.

PR #373, opened 2022-09-02
filter out instances in different vpc networks

tyuchn:
**What type of PR is this?**
>
> /kind api-change
kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

filter out instances in different vpc networks at ip reservation step

Fixes #b/244645541

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -281,6 +281,9 @@ func (s *controllerServer) getCloudInstancesReservedIPRanges(ctx context.Context
 	if err != nil {
 		return nil, status.Error(codes.Aborted, err.Error())
 	}
+	// Due to unreachable location some instances may not show up here.
+	// TODO: create a new function to take a list of locations
+	// and return error if unreachable contained the region of interest.
 	multiShareInstances, err := s.config.fileService.ListMultishareInstances(ctx, &file.ListFilter{Project: filer.Project, Location: "-"})
 	if err != nil {
 		return nil, status.Error(codes.Aborted, err.Error())
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -286,13 +286,19 @@ func (s *controllerServer) getCloudInstancesReservedIPRanges(ctx context.Context
 		return nil, status.Error(codes.Aborted, err.Error())
 	}
 
-	// Initialize an empty reserved list. It will be populated with all the reservedIPRanges obtained from the cloud instances
+	// Initialize an empty reserved list. It will be populated with all the
+	// reservedIPRanges obtained from the cloud instances in the same VPC network
+	// as the ServiceInstance.
 	cloudInstancesReservedIPRanges := make(map[string]bool)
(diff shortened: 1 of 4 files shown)


saikat-royc:
Please ensure to test the PR manually (if not already done)

saikat-royc:
Can you also add some unit test for this?

saikat-royc:
Overall change LGTM

tyuchn:
> 

Done

tyuchn:
> Can you also add some unit test for this?

Done

pkg/csi_driver/controller.go
```diff
@@ -281,18 +281,27 @@ func (s *controllerServer) getCloudInstancesReservedIPRanges(ctx context.Context
...
 	}
 
-	// Initialize an empty reserved list. It will be populated with all the reservedIPRanges obtained from the cloud instances
+	// Initialize an empty reserved list. It will be populated with all the
+	// reservedIPRanges obtained from the cloud instances in the same VPC network
+	// as the ServiceInstance.
 	cloudInstancesReservedIPRanges := make(map[string]bool)
 	for _, instance := range instances {
-		cloudInstancesReservedIPRanges[instance.Network.ReservedIpRange] = true
+		if strings.EqualFold(instance.Network.Name, filer.Network.Name) {
```
leiyiz, line 297: what would happen for default Network here? i.e. if the customer does not specify any network when creating instances, would the `filer.Network.Name` and `instance.Network.Name` still match up?

pkg/csi_driver/controller_test.go
```diff
@@ -907,3 +907,73 @@ func TestCreateSnapshot(t *testing.T) {
...
+				Project:  testProject,
+				Name:     testCSIVolume,
+				Location: testLocation,
+				Tier:     defaultTier,
+				Network: file.Network{
+					Name:        defaultNetwork,
+					ConnectMode: directPeering,
+				},
+			},
+			expectIPRange: map[string]bool{"192.168.92.32/29": true, "192.168.92.40/29": true, "10.1.1.0/24": true},
```
leiyiz, line 963: why would there be `192.168.92.32/29` and `192.168.92.40/29`?

pkg/csi_driver/controller.go
tyuchn: If network is not specified, `default` will be stamped as the network value. 
See https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/controller.go#L453

pkg/csi_driver/controller_test.go
tyuchn: [Here's](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/cloud_provider/file/fake.go#L119-L143) where they are defined. A fake `ListInstances` will be called from the fake controller during the test.

leiyiz (commented):
/approve

PR #376, opened 2022-09-30
set default instance size for enterprise

amacaskill:
**What type of PR is this?**
> /kind feature

This PR sets a default instance size for filestore enterprise tier. We currently [set a minimum instance size of 1Tb](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/controller.go#L37), for the the standard tier, so customers will not see any issues even if they specify a storage capacity smaller than the minimum. We do not do this for enterprise tier which can lead to customer issues where volumes cannot be created because the instance size is smaller than the minimum. We should be consistent with both tiers. 

Fixes #

```release-note
Minimum instance size of 2.5Tb is set for enterprise tier volumes. 
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -34,10 +34,10 @@ import (
 )
 
 const (
-	// premium tier min is 2.5 Tb, let GCFS error
-	minVolumeSize     int64 = 1 * util.Tb
-	modeInstance            = "modeInstance"
-	newInstanceVolume       = "vol1"
+	minVolumeSizeStandard   int64 = 1 * util.Tb
+	minVolumeSizeEnterprise int64 = 2.5 * util.Tb
+	modeInstance                  = "modeInstance"
+	newInstanceVolume             = "vol1"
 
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/controller.go
```diff
@@ -34,10 +34,10 @@ import (
 )
 
 const (
-	// premium tier min is 2.5 Tb, let GCFS error
-	minVolumeSize     int64 = 1 * util.Tb
-	modeInstance            = "modeInstance"
-	newInstanceVolume       = "vol1"
+	minVolumeSizeStandard   int64 = 1 * util.Tb
+	minVolumeSizeEnterprise int64 = 2.5 * util.Tb
```
leiyiz, line 38: the minimum for enterprise tier is 1T and the minimum for premium tier is 2.5T.

leiyiz (commented):
I thought we have talked about this and @mattcary said it's not good to silently provision 2.5T of storage when user is expecting 20

PR #380, opened 2022-11-11
Fix of driver error wrapping hides error codes

sunnylovestiramisu:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

We discovered a bug in the driver that causes the wrong error code to be reported. glog.Errorf does not support %w and we need to unwrap the error to identify apiErr.

Fixes #

NONE

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ dynamically created and mounted by workloads.
 ## Project Status
 Status: GA
 
-Latest image: `k8s.gcr.io/cloud-provider-gcp/gcp-filestore-csi-driver:v1.3.4`
+Latest image: `k8s.gcr.io/cloud-provider-gcp/gcp-filestore-csi-driver:v1.3.5`
 
 Also see [known issues](KNOWN_ISSUES.md) and [CHANGELOG](CHANGELOG.md).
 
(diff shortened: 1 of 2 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -31,7 +32,7 @@ import (
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
```
leiyiz, line 35: hmm I wonder if moving to v2 will affect other logging as well. V1 is deprecated so we SHOULD be using v2 but is v2 backwards compatible? @msau42 do you know?

msau42: The Verbose type changed which caused klog to rev to v2: https://github.com/kubernetes/klog/commit/cd60aa438f9c750641d5feacd2999e58df8af28f

This may also be important if we are supporting both klog v1 AND v2 AND glog? We probably want to convert it all https://github.com/kubernetes/klog#coexisting-with-klogv2

PR #381, opened 2022-11-11
Fix of driver error wrapping hides error codes

sunnylovestiramisu:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

We discovered a bug in the driver that causes the wrong error code to be reported. glog.Errorf does not support %w and we need to unwrap the error to identify apiErr.

Fixes #

```release-note
NONE
```


pkg/cloud_provider/file/file.go
```diff
@@ -31,7 +32,7 @@ import (
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
```
leiyiz, line 35: hmm I wonder if moving to v2 will affect other logging as well. V1 is deprecated so we SHOULD be using v2 but is v2 backwards compatible? @msau42 do you know?

leiyiz: moving this comment from #380 because it's closed

leiyiz (commented):
/ok-to-test

pkg/cloud_provider/file/file.go
msau42: The Verbose type changed which caused klog to rev to v2: https://github.com/kubernetes/klog/commit/cd60aa438f9c750641d5feacd2999e58df8af28f

This may also be important if we are supporting both klog v1 AND v2 AND glog? We probably want to convert it all https://github.com/kubernetes/klog#coexisting-with-klogv2

leiyiz: IIUC: anywhere we use klog.V() might be broken because klog v2 changed this to return a struct instead of a bool? Tests have been passing so sounds like mostly we don't use that

sunnylovestiramisu: I did not see we use the return value of klog.V() in the code base so it should not impact anything. Do you have know if there is any place we might be doing that?

leiyiz: yeah that's my understanding. should be the same. code search don't show up anything

leiyiz (commented):
lgtm in general. please squash the commits and we can merge it in.

leiyiz (commented):
/lgtm

pkg/cloud_provider/file/file.go
msau42: Also double check that the various types of logs, info, warn, error, etc are still printed as expected. There was some incompatibility between command line arguments between glog, klog and klog v2

leiyiz: @sunnylovestiramisu can you double check as suggested

msau42: See https://github.com/kubernetes-csi/external-provisioner/pull/518 for an example issue that can happen if you have multiple klog versions

cmd/webhook/main.go
```diff
@@ -19,7 +19,7 @@ package main
 import (
 	"flag"
 
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
```
msau42, line 22: I would expect that klog and glog can be removed from go.mod. Can you run "go mod tidy" and see if that cleans it up?

msau42, line 22: There may also be some differences in how we iniitalize flags between glog and klog. See https://github.com/kubernetes-csi/external-provisioner/pull/225/files for an example.

sunnylovestiramisu: I just ran ``go mod tidy`` but no new changes show up.

sunnylovestiramisu: Searched flag.CommandLine in the this package nothing came up. Do we need to do this proactively?

sunnylovestiramisu: Double checked that we used commanline flags in  setup_e2e_test.go but it is already doing klog.InitFlags(flag.CommandLine).

pkg/cloud_provider/file/file.go
sunnylovestiramisu: From the test logs: http://shortn/_e4Bnk4egqd

```
klog.Errorf() in file.go:318 Failed to get...
klog.Infof() in controller.go:260 CreateVolume succeeded...
```

From pod logs: http://shortn/_ONYnRuhl5B
```
klog.Warningf() in mount_helper_common.go:133] Warning...
```

pkg/csi_driver/controller.go
```diff
@@ -128,7 +127,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		return response, err
 	}
 
-	glog.V(4).Infof("CreateVolume called with request %+v", req)
+	klog.V(4).Infof("CreateVolume called with request %+v", req)
```
msau42, line 130: It looks like log messages that specify specific verbosity is no longer being printed:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-filestore-csi-driver/381/pull-gcp-filestore-csi-driver-kubernetes-integration/1592940958024994816/artifacts/gcp-fs-csi-driver/gcp-filestore-csi-controller-86456464b6-zx9sr-gcp-filestore-driver.log

Log from a different PR:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-filestore-csi-driver/371/pull-gcp-filestore-csi-driver-kubernetes-integration/1565104360969474048/artifacts/gcp-fs-csi-driver/gcp-filestore-csi-controller-78f8ddcc6-gtpmz-gcp-filestore-driver.log

cmd/webhook/main.go
msau42: I think we need to add `klog.InitFlags()` here: https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/cmd/main.go#L57

sunnylovestiramisu: Added.

msau42: Can you run `go mod vendor` and `go mod tidy` again? I would expect that the entries under go.sum and https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/tree/master/vendor/github.com/golang/glog to be removed.

pkg/csi_driver/controller.go
sunnylovestiramisu: It showed up in the latest run: http://shortn/_kmyhJUtOvN

```
I1117 23:40:02.397994       1 controller.go:130] CreateVolume called with request name:"pvc-10587a49-55aa-4c72-a7a0-c6dbd142f026" capacity_range:<required_bytes:1099511627776 > volume_capabilities:<mount:<> access_mode:<mode:SINGLE_NODE_WRITER > > parameters:<key:"csi.storage.k8s.io/pv/name" value:"pvc-10587a49-55aa-4c72-a7a0-c6dbd142f026" > parameters:<key:"csi.storage.k8s.io/pvc/name" value:"csi-gcpfs-fs-sc-basic-hddvxcfw" > parameters:<key:"csi.storage.k8s.io/pvc/namespace" value:"volume-expand-4087" > parameters:<key:"network" value:"csi-filestore-test-network" > accessibility_requirements:<requisite:<segments:<key:"topology.gke.io/zone" value:"us-central1-b" > > preferred:<segments:<key:"topology.gke.io/zone" value:"us-central1-b" > > > 
```

cmd/webhook/main.go
sunnylovestiramisu: The weird thing is that ```go mod tidy``` always try to add k8s.io/klog v1.0.0 back. :(

msau42: It probably means that one of the depenencies is still using klog v1

PR #390, opened 2022-11-18
Silent volume provisioning surprise

hsadoyan:
What type of PR is this?

Uncomment only one /kind <> line, hit enter to put that in a new line, and remove leading whitespaces from that line:

/kind bug

What this PR does / why we need it:
See Below

Which issue(s) this PR fixes:

Fixes #

Special notes for your reviewer:

Does this PR introduce a user-facing change?:

```release-note
Previous Behavior: 
   * if a user requested a volume size smaller than the Standard minimum (1TB), we'd quietly round up to the standard Minimum.
   * we didn't account for different tiers having different minimums
   * we didn't check against the maximum volume size for a tier

New Behavior: 
  * Compare against min/max values of provided tier, instead of always standard. (If tier is unrecognized, default to standard)
  * If the requested min is less than possible min, and no upper bound is given, return an error instead of quietly rounding up
  * vice versa for max
  
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -33,14 +33,25 @@ import (
 )
 
 const (
-	// premium tier min is 2.5 Tb, let GCFS error
-	minVolumeSize     int64 = 1 * util.Tb
-	modeInstance            = "modeInstance"
-	newInstanceVolume       = "vol1"
+	modeInstance      = "modeInstance"
+	newInstanceVolume = "vol1"
 
 	defaultTier    = "standard"
 	enterpriseTier = "enterprise"
+	premiumTier    = "premium"
+	basicHDDTier   = "basic_hdd"
+	basicSSDTier   = "basic_ssd"
+	highScaleTier  = "high_scale_ssd"
 	defaultNetwork = "default"
+
+	defaultTierMinSize    = 1 * util.Tb
+	defaultTierMaxSize    = 639 * util.Tb / 10
+	enterpriseTierMinSize = 1 * util.Tb
+	enterpriseTierMaxSize = 10 * util.Tb
+	highScaleTierMinSize  = 10 * util.Tb
+	highScaleTierMaxSize  = 100 * util.Tb
+	premiumTierMinSize    = 25 * util.Tb / 10
+	premiumTierMaxSize    = 639 * util.Tb / 10
 
 	directPeering        = "DIRECT_PEERING"
 	privateServiceAccess = "PRIVATE_SERVICE_ACCESS"
@@ -83,6 +94,11 @@ const (
 	tagKeyClusterName              = "storage_gke_io_cluster_name"
 	tagKeyClusterLocation          = "storage_gke_io_cluster_location"
 )
+
+type capacityRangeForTier struct {
+	min int64
+	max int64
+}
 
 // controllerServer handles volume provisioning
 type controllerServer struct {
@@ -137,7 +153,8 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	capBytes, err := getRequestCapacity(req.GetCapacityRange())
+	tier := getTierFromParams(req.GetParameters())
+	capBytes, err := getRequestCapacity(req.GetCapacityRange(), tier)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/controller.go
```diff
@@ -417,36 +430,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
-		return 0, fmt.Errorf("limit bytes %v is less than required bytes %v", lCap, rCap)
+	minRequired := capRange.GetRequiredBytes()
+	minSet := minRequired > 0
+	maxRequested := capRange.GetLimitBytes()
+	maxSet := maxRequested > 0
+
+
+
+	if maxSet && minSet && maxRequested < minRequired {
+		return fmt.Errorf("limit bytes %v is less than required bytes %v", maxRequested, minRequired)
```
tyuchn, line 453: I think we should use %d to print int64 here and below.

pkg/csi_driver/controller.go
```diff
@@ -417,36 +437,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
-	rSet := rCap > 0
-	lCap := capRange.GetLimitBytes()
-	lSet := lCap > 0
+	return defaultTier
+}
+
+func invalidCapacityRange(capRange *csi.CapacityRange, tier string)  error {
+	validRange := provisionableCapacityForTier(tier)
+
+	minRequired := capRange.GetRequiredBytes()
```
leiyiz, line 452: consider change this to `requiredCap` to align with naming in `CapacityRange`

pkg/csi_driver/controller.go
```diff
@@ -417,36 +437,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
+	minSet := minRequired > 0
+	maxRequested := capRange.GetLimitBytes()
```
leiyiz, line 454: change this to `limitCap`

pkg/csi_driver/controller.go
```diff
@@ -417,36 +437,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
-	lCap := capRange.GetLimitBytes()
-	lSet := lCap > 0
+	return defaultTier
+}
+
+func invalidCapacityRange(capRange *csi.CapacityRange, tier string)  error {
+	validRange := provisionableCapacityForTier(tier)
+
+	minRequired := capRange.GetRequiredBytes()
+	minSet := minRequired > 0
```
leiyiz, line 453: consider change to `requireSet`

pkg/csi_driver/controller.go
```diff
@@ -417,36 +437,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
+	maxRequested := capRange.GetLimitBytes()
+	maxSet := maxRequested > 0
```
leiyiz, line 455: `limitSet`

pkg/csi_driver/controller.go
```diff
@@ -417,36 +437,99 @@ func (s *controllerServer) ControllerGetCapabilities(ctx context.Context, req *c
...
+		}
+
+		if !maxSet && minRequired < validRange.min { 
+			// Avoid surprising users by provisioning more than Requested
+			return fmt.Errorf("limit bytes %v is less than minimum instance size bytes %v for tier %s, but no maximum was specified", minRequired, validRange.min, tier)
+		}
 	}
+	if maxSet {
+		if maxRequested < validRange.min {
+			return fmt.Errorf("limit bytes %v is less than minimum instance size bytes %v for tier %s", maxRequested, validRange.min, tier)
```
leiyiz, line 475: if we try to surface these message to users, do we want to make them clear by saying in terms of`TiB` instead of bytes?

leiyiz:
please add a release note section in your PR description too. like so:

``` release-note
xxxxx
```

pkg/csi_driver/controller.go
hsadoyan: Good point! I followed the existing messages, but the TiB values would be much more legible.

pkg/csi_driver/controller.go
hsadoyan: Done.

pkg/csi_driver/controller.go
```diff
@@ -446,59 +443,59 @@ func getTierFromParams(params map[string]string) string {
...
+	if limitSet && requireSet && limitCap < requiredCap {
+		return fmt.Errorf("limit bytes %v is less than required bytes %v", limitCap, requiredCap)
 	}
 
-	if minSet{
-		if minRequired > validRange.max {
-			return fmt.Errorf("request bytes %v is more than maximum instance size bytes %v for tier %s", minRequired, validRange.max, tier)
+	if requireSet {
+		if requiredCap > validRange.max {
+			return fmt.Errorf("request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s", requiredCap/util.Tb, validRange.max/util.Tb, tier)
```
leiyiz, line 461: `requiredCap/util.Tb` will round to an int IIUC, we might be returning incorrect info.

PR #399, opened 2022-12-22
tests: add test for getting region from zone

ajayk:
**What type of PR is this?**

 /kind cleanup

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -143,14 +143,14 @@ func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
 	return labelsMap, nil
 }
 
+// GetRegionFromZone return the corresponding region name based on a zone name
+// Example "us-central1-a" return "us-central1"
 func GetRegionFromZone(location string) (string, error) {
 	tokens := strings.Split(location, "-")
 	if len(tokens) != 3 {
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -143,14 +143,14 @@ func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
...
 func GetRegionFromZone(location string) (string, error) {
 	tokens := strings.Split(location, "-")
 	if len(tokens) != 3 {
-		return "", fmt.Errorf("Failed to parse location %v", location)
+		return "", fmt.Errorf("failed to parse location %v", location)
 	}
-
-	tokens = tokens[:2]
-	return strings.Join(tokens, "-"), nil
+	return strings.Join(tokens[0:len(tokens)-1], "-"), nil
```
leiyiz, line 153: is it needed here to do `len(tokens)-1` when we already know that the length has to be 3?

ajayk: i don't i think its needed ,probably this would make it better `strings.Join(token[0:2],"-")` ?

leiyiz (commented):
/lgtm
/approve

PR #404, opened 2023-01-10
Improve messaging for common failures

hsadoyan:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Long running operations (particularly around multi-share) often get retried while the initial operation is still running. The retry attempt fails, leading to confusing error messages in the pvc events.

Example: 
``` 
 Warning  ProvisioningFailed    4m10s                  filestore.csi.storage.gke.io_gke-uuid failed to provision volume with StorageClass "enterprise-multishare-rwx": rpc error: code = Aborted desc = An operation with the given volume key pvc-foo-bar already exists. 
 ```
 
 
The provisioning did not, in fact, fail. Only the current retry was canceled early, since the previous attempt is still running. Unfortunately that status is set by the caller, so we can't change it here, so we chose to amend the common error messages to help explain what might be happening. 

Screenshot 2023-01-10 at 5 07 20 PM

There is still a wall of other text added by the caller before the event is created, but the description now offers more context.  

The other option was to filter these out entirely, but there is still benefit to providing the extra information to the user.

```release-note
Improve error messaging during common retries. 
```

### DIFF
--- a/pkg/csi_driver/multishare_ops_manager.go
+++ b/pkg/csi_driver/multishare_ops_manager.go
@@ -99,7 +99,7 @@ func (m *MultishareOpsManager) setupEligibleInstanceAndStartWorkflow(ctx context
 	}
 
 	// No share or running share create op found. Proceed to eligible instance check.
-	eligible, numIneligible, err := m.runEligibleInstanceCheck(ctx, req, ops, instance, regions)
+	eligible, err := m.runEligibleInstanceCheck(ctx, req, ops, instance, regions)
 	if err != nil {
 		return nil, nil, status.Error(codes.Aborted, err.Error())
 	}
@@ -126,11 +126,6 @@ func (m *MultishareOpsManager) setupEligibleInstanceAndStartWorkflow(ctx context
 
 		w, err := m.startShareWorkflow(ctx, &Workflow{share: share, opType: util.ShareCreate}, ops)
 		return w, nil, err
-	}
-
-	if numIneligible > 0 {
-		// some instances not ready yet. wait for more instances to be ready.
-		return nil, nil, status.Errorf(codes.Aborted, " %d non-ready instances detected. No ready instance found. \n --- Most likely a long process is still running to completion. Retrying.", numIneligible)
 	}
 
 	param := req.GetParameters()
@@ -344,11 +339,11 @@ func (m *MultishareOpsManager) verifyNoRunningInstanceOrShareOpsForInstance(inst
 }
 
 // runEligibleInstanceCheck returns a list of ready and non-ready instances.
-func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req *csi.CreateVolumeRequest, ops []*OpInfo, target *file.MultishareInstance, regions []string) ([]*file.MultishareInstance, int, error) {
+func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req *csi.CreateVolumeRequest, ops []*OpInfo, target *file.MultishareInstance, regions []string) ([]*file.MultishareInstance, error) {
 	klog.Infof("ListMultishareInstances call initiated for request %+v.", req)
 	instances, err := m.listMatchedInstances(ctx, req, target, regions)
 	if err != nil {
-		return nil, 0, err
+		return nil, err
 	}
 	klog.Infof("ListMultishareInstances call returned successfully with %d instances for request %+v.", len(instances), req)
 	// An instance is considered as eligible if and only if the state is 'READY', and there's no ops running against it.
@@ -374,14 +369,14 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req
 		op, err := containsOpWithInstanceTargetPrefix(instance, ops)
 		if err != nil {
 			klog.Errorf("failed to check eligibility of instance %s", instance.Name)
-			return nil, 0, err
+			return nil, err
 		}
 
 		if op == nil {
 			shares, err := m.cloud.File.ListShares(ctx, &file.ListFilter{Project: instance.Project, Location: instance.Location, InstanceName: instance.Name})
 			if err != nil {
 				klog.Errorf("Failed to list shares of instance %s/%s/%s, err:%v", instance.Project, instance.Location, instance.Name, err.Error())
-				return nil, 0, err
+				return nil, err
 			}
 			if len(shares) >= util.MaxSharesPerInstance {
 				continue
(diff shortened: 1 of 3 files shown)


saikat-royc:
This is not sufficient for the user to understand that a potential long share delete is blocking new volume (aka share) creation.

So there needs to be additional details around the error string "x non-ready instances detected. No ready instance found".

We capture the operation type and id which marks the instance as non-ready here https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/d447319506553a613e62b6dade7a047ad357efe9/pkg/csi_driver/multishare_ops_manager.go#L395
This log is the most important as it highlights that CSI did not find any ready instance to initiate a new share create.

Can you experiment with how does it (kubectl events) if we emit event for the given PVC key to capture a string like instance x busy with long running operation y of type z.  similar to how provisioner does it [code](https://github.com/kubernetes-csi/external-provisioner/blob/831b9c5d80374f673f16bcd615f40f5a832ed0bf/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v8/controller/controller.go#L1412)

We already have the claim details in the csi createVolumeRequest https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/d447319506553a613e62b6dade7a047ad357efe9/pkg/csi_driver/multishare_ops_manager.go#L347; https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/d447319506553a613e62b6dade7a047ad357efe9/pkg/csi_driver/multishare_controller.go#L497

hsadoyan:
> We capture the operation type and id which marks the instance as non-ready here
> 
> https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/d447319506553a613e62b6dade7a047ad357efe9/pkg/csi_driver/multishare_ops_manager.go#L395
> 
> 
> This log is the most important as it highlights that CSI did not find any ready instance to initiate a new share create.
> Can you experiment with how does (kubectl events) show up , if we emit event for the given PVC key to capture a string like instance x busy with long running operation y of type z. similar to how provisioner does it [code](https://github.com/kubernetes-csi/external-provisioner/blob/831b9c5d80374f673f16bcd615f40f5a832ed0bf/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v8/controller/controller.go#L1412)

If a customer has multiple multishare instances, we'll try to find at least one eligible one. I'm worried that if send out an event like that for every instance, things will be even more cluttered, since they'll see "instance x busy with long running operation y of type z" messages alongside messages they already get. And if there is 1 ready instance and x non-ready instances, then the user would still see the messages about the non-ready instances even though we were able to take the ready one. 

Instead of listing all the running operations, could we do something like `All available instances are currently busy. Waiting...`

hsadoyan:
We can also list all the Non-ready instances and their operations, but only in the case where there are no ready instances. Something like: 

"No ready instances found. 3 non-ready instances found. 

Instance x busy with operation y of type z.
Instance x1 busy with operation y1 of type z.
Instance x2 busy with operation y2 of type z.

Waiting for free instance."

hsadoyan:
Screenshot 2023-01-23 at 11 09 26 AM


Updated PR. Here's an example of how it'll look like with a delete in progress.

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -393,11 +388,26 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req
...
+	if len(readyEligibleInstances) == 0 && len(nonReadyEligibleInstances) > 0 {
+		errorString := "All eligible instances are busy.\n"
+
+		for _, instance := range nonReadyEligibleInstances {
+			op, _ := containsOpWithInstanceTargetPrefix(instance, ops) // Error for this call is already checked above
+			errorString = fmt.Sprintf("%s Instance %s busy with operation type %s\n", errorString, instance.Name, op.Type)
+
+		}
+
+		errorString = errorString + "Waiting for instance..."
```
saikat-royc, line 405: + "Waiting for instance..." is not needed I think.

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -393,11 +388,26 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req
...
 		klog.Infof("Instance %s/%s/%s with state %s is not ready with ongoing operation %s type %s", instance.Project, instance.Location, instance.Name, instance.State, op.Id, op.Type.String())
-		nonReadyInstanceCount += 1
+		nonReadyEligibleInstances = append(nonReadyEligibleInstances, instance)
+
 		// TODO: If we see > 1 instances with 0 shares (these could be possibly leaked instances where the driver hit timeout during creation op was in progress), should we trigger delete op for such instances? Possibly yes. Given that instance create/delete and share create/delete is serialized, maybe yes.
 	}
 
-	return readyEligibleInstances, nonReadyInstanceCount, nil
+	if len(readyEligibleInstances) == 0 && len(nonReadyEligibleInstances) > 0 {
+		errorString := "All eligible instances are busy.\n"
```
saikat-royc, line 397: All eligible filestore instances

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -393,11 +388,26 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req
...
+
+		for _, instance := range nonReadyEligibleInstances {
+			op, _ := containsOpWithInstanceTargetPrefix(instance, ops) // Error for this call is already checked above
+			errorString = fmt.Sprintf("%s Instance %s busy with operation type %s\n", errorString, instance.Name, op.Type)
```
saikat-royc, line 401: for the scenario of line 360 above, where instance state != READY, I think its better report Instance is in "x" state, instead of ops.

For instance.state == READY, we can report busy with operation {opid}

Thoughts?

hsadoyan: Good point. Changed it to report state if OP isn't available.

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -393,11 +388,32 @@ func (m *MultishareOpsManager) runEligibleInstanceCheck(ctx context.Context, req
...
 
-	return readyEligibleInstances, nonReadyInstanceCount, nil
+	if len(readyEligibleInstances) == 0 && len(nonReadyEligibleInstances) > 0 {
+		errorString := "All eligible filestore instances are busy.\n"
+
+		for _, instance := range nonReadyEligibleInstances {
+			op, err := containsOpWithInstanceTargetPrefix(instance, ops) // Error for this call is already checked above
+			if err != nil {
+				klog.Errorf("failed to check eligibility of instance %s", instance.Name)
+				return nil, err
```
saikat-royc, line 403: by the time we have reached line 396, we have already determined the `readyEligibleInstances` and `nonReadyEligibleInstances`. Can we just log a kWarning here and continue?

hsadoyan: If we log a klog warning, the user won't be able to see it in their events.  

The if statement here is to keep the same decision tree as we had before. Before the caller method would check if there are no ready instances, and nonReady was >0 it would return the error event for user to see. (Previously line 131)

I didn't want to change the actual logic of when we show the error.

saikat-royc: Ack

PR #417, opened 2023-02-02
return error code from waitOnWorklow for multishare driver

amacaskill:
**What type of PR is this?**
> /kind cleanup

Right now we return an Internal on CreateVolume for filestore create/delete instance /share. The filestore instance create/delete takes longer than 5 minutes (around 10 min), so we are always getting failures that count against our SLO even  though this is intended. After this is submitted, I will filter out DeadlineExceeded from our SLO.  

Manually tested for CreateVolume: 
- `Canceled`: https://screenshot.googleplex.com/3AvQajefdnYot6w.png
  - We get `Canceled` when the external provisioner cancels the Context with [cancel()](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L796)
- `DeadlineExceeded`: https://screenshot.googleplex.com/6gxJpn6JFdYpDwm.png
  - `DeadlineExceeded` is returned by wait.Poll when the [wait.Poll](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/cloud_provider/file/file.go#L865) times out.  
- `ResourceExhausted`: https://screenshot.googleplex.com/7To8W8B8hLK6aVn.png  
  - I am actually getting Internal AND `ResourceExhausted` from the same source location: [utils.go line 59](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/release-1.3/pkg/csi_driver/utils.go#L59) https://screenshot.googleplex.com/6di7QpcEfi4S6AC.png but the slo is based on the return code that the sidecar receives, so these Internal error codes should not affect the SLO. 
 

```release-note
Return DeadlineExceeded / Canceled or respective user error code instead of Internal error code when the create/delete filestore instance/share context times out / gets canceled or encounters a user error during polling.
```

### DIFF
--- a/pkg/csi_driver/multishare_controller.go
+++ b/pkg/csi_driver/multishare_controller.go
@@ -113,7 +113,8 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
 	// lock released. poll for op.
 	err = m.waitOnWorkflow(ctx, workflow)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Create Volume failed, operation %q poll error: %v", workflow.opName, err)
+		errCode := file.PollOpErrorCode(err)
+		return nil, status.Errorf(*errCode, "Create Volume failed, operation %q poll error: %v", workflow.opName, err)
 	}
 
 	klog.Infof("Poll for operation %s (type %s) completed", workflow.opName, workflow.opType.String())
@@ -140,7 +141,8 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
 	// lock released. poll for share create op.
 	err = m.waitOnWorkflow(ctx, shareCreateWorkflow)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "%s operation %q poll error: %v", shareCreateWorkflow.opType.String(), shareCreateWorkflow.opName, err)
+		errCode := file.PollOpErrorCode(err)
+		return nil, status.Errorf(*errCode, "%s operation %q poll error: %v", shareCreateWorkflow.opType.String(), shareCreateWorkflow.opName, err)
 	}
 	return m.getShareAndGenerateCSICreateVolumeResponse(ctx, instanceScPrefix, newShare)
 }
@@ -199,7 +201,8 @@ func (m *MultishareController) DeleteVolume(ctx context.Context, req *csi.Delete
 	if workflow != nil {
 		err = m.waitOnWorkflow(ctx, workflow)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "%s operation %q poll error: %v", workflow.opType.String(), workflow.opName, err.Error())
+			errCode := file.PollOpErrorCode(err)
+			return nil, status.Errorf(*errCode, "%s operation %q poll error: %v", workflow.opType.String(), workflow.opName, err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


amacaskill:
Manually tested for CreateVolume: https://screenshot.googleplex.com/4tTRiWLy3vtG9sb.png

pkg/csi_driver/multishare_controller.go
```diff
@@ -294,9 +297,10 @@ func (m *MultishareController) ControllerExpandVolume(ctx context.Context, req *
 		return nil, err
 	}
 
-	err = m.waitOnWorkflow(ctx, workflow)
+	err, errCode := m.waitOnWorkflow(ctx, workflow)
+	// Filter out these. Add progressdeadlineexceeded.
```
saikat-royc, line 301: we can remove this comment?

pkg/csi_driver/multishare_controller.go
```diff
@@ -334,13 +339,18 @@ func (m *MultishareController) getShareAndGenerateCSIControllerExpandVolumeRespo
...
 
-func (m *MultishareController) waitOnWorkflow(ctx context.Context, workflow *Workflow) (err error) {
+func (m *MultishareController) waitOnWorkflow(ctx context.Context, workflow *Workflow) (error, *codes.Code) {
 	timeout, pollInterval, err := util.GetMultishareOpsTimeoutConfig(workflow.opType)
 	if err != nil {
-		return
+		code := codes.Internal
+		return err, &code
 	}
 	err = m.cloud.File.WaitForOpWithOpts(ctx, workflow.opName, file.PollOpts{Timeout: timeout, Interval: pollInterval})
```
saikat-royc, line 348: For the error returned in the poll operation, we should have a function for resource exhausted (the IsUserError() handles that), as we can potentially hit that. We do not want to mask those errors as deadline exceeded.

pkg/csi_driver/multishare_controller.go
amacaskill: Yes removed

pkg/csi_driver/multishare_controller.go
amacaskill: So you are saying in WaitForOpWithOpts, this line
[pollOp, err := manager.multishareOperationsServices.Get(op).Context(ctx).Do()](
https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/ad542caf0e1276e49016b3c33735de4515e8e318/pkg/cloud_provider/file/file.go#L866) can return ResourceExhausted? Looking at the definition of Do(), it also wraps the errors it returns: https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/ad542caf0e1276e49016b3c33735de4515e8e318/vendor/google.golang.org/api/file/v1beta1/file-gen.go#L6490

saikat-royc: we can occasionally get 404 too, if the resource disappeared between poll interval I think

amacaskill: Hm ok I so return Internal error code for anything returned from `GetMultishareOpsTimeoutConfig()` or [IsOpDone()](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/ad542caf0e1276e49016b3c33735de4515e8e318/pkg/cloud_provider/file/file.go#L593). 

Then for  Do() we need to map 404,  resource exhausted, and deadline exceeded to the correct error code, then for all others we return an internal error code. The polling error that has the "context deadline exceeded" message isn't actually a googleapi error right? Should I just parse the error string to see if it has that string, if so return DeadlineExceeded, otherwise return Internal?

saikat-royc: Yes thats sounds a reasonable approach

saikat-royc:
/lgtm
/approve

Thanks for fixing this!

PR #446, opened 2023-03-08
stateful csi driver custom resource manipulation logic

leiyiz:
**What type of PR is this?**
/kind feature

This PR contains reconciler logic for reconciling the custom resource object.

Fixes #

`main.go` is where the reconciler is launched
`reconciler.go` contains the core logic of the reconciler from [reconstructing CRs](https://docs.google.com/document/d/14czaZ-cT5F60YLd9rXw8BrmlaoYzlCi5zMs3rZg0xdI/edit?pli=1#bookmark=id.t6vhhgpsosrf) to [updating CRs](https://docs.google.com/document/d/14czaZ-cT5F60YLd9rXw8BrmlaoYzlCi5zMs3rZg0xdI/edit?pli=1#bookmark=id.486ksm8vt2tu).
The rest of the reconciler is going to be in another PR.

- the `shareInfo` object's name is equal to the PVName with "-" and the `shareInfo.Spec.ShareName` is the actual share name on the filestore side with "_"
- `instanceInfo.Name` is full `instanceUri` with "/" replaced by "." and `instanceInfo.Status.ShareNames` contains `shareInfo.Name` with "-"
- unit testing the reconciler will come later as it is going to be a bug task to create test framework to test state transitions

/assign @tyuchn 
/assign @saikat-royc 

```release-note
NONE
```


cmd/main.go
```diff
@@ -53,11 +67,15 @@ var (
 	leaderElectionRetryPeriod   = flag.Duration("leader-election-retry-period", 5*time.Second, "Duration, in seconds, the LeaderElector clients should wait between tries of actions. Defaults to 5 seconds.")
 	lockReleaseSyncPeriod       = flag.Duration("lock-release-sync-period", 60*time.Second, "Duration, in seconds, the sync period of the lock release controller. Defaults to 60 seconds.")
 
+	enableStateful = flag.Bool("enable-stateful-multishare", false, "")
```
saikat-royc, line 70: we are following a standard prefix "feature-". e.g https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/cmd/main.go#L49 for feature-lock-release

cmd/main.go
```diff
@@ -129,6 +147,10 @@ func main() {
 		FeatureOptions:   featureOptions,
 	}
 
+	if *runController && *enableMultishare && *enableStateful {
+		runMultishareReconciler(config)
```
saikat-royc, line 151: can we move this reconciler inside the [GCFSDriver.run()](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/gcfs_driver.go#L228) to keep the reconciler controllers in one place?

cmd/main.go
```diff
@@ -139,3 +161,101 @@ func main() {
...
+			klog.Errorf("Failed to list v1alpha1 shareinfos with error=%+v", err)
+			return false, nil
+		}
+
+		return true, nil
+	}
+
+	// The maximum retry duration = initial duration * retry factor ^ # steps. Rearranging, this gives
+	// # steps = log(maximum retry / initial duration) / log(retry factor).
+	const retryFactor = 1.5
```
saikat-royc, line 234: please create some well defined constants for these

cmd/main.go
```diff
@@ -139,3 +161,101 @@ func main() {
...
+
+	recon := reconciler.NewMultishareReconciler(
+		fsClient,
+		driverConfig,
+		factory.Multishare().V1alpha1().ShareInfos(),
+		factory.Multishare().V1alpha1().InstanceInfos(),
+		coreFactory.Storage().V1().StorageClasses().Lister(),
+	)
+
+	if err := ensureCustomResourceDefinitionsExist(fsClient); err != nil {
```
saikat-royc, line 195: if we move the runMultishareReconciler() to GCFSDriver.run(), the ensure CRD  exists can be moved out and remain in cmd/main.go under the `*runController && *enableMultishare && *enableStateful` check. Thus by the time we reach the reconciler start logic, we are ensured the CRD definition exits.

Thoughts on this restructure?

pkg/multishare_reconciler/reconciler_test.go
```diff
@@ -0,0 +1,297 @@
+package multishare_reconciler
```
saikat-royc, line 1: add copyright notice

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	klog.Infof("Starting multishare reconciler")
+	defer klog.Infof("Shutting down multishare reconciler")
+
+	informerSynced := []cache.InformerSynced{recon.shareListerSynced, recon.instanceListerSynced}
+	if !cache.WaitForCacheSync(stopCh, informerSynced...) {
+		klog.Errorf("Cannot sync caches")
+		return
+	}
+
+	go wait.Until(recon.reconcileWorker, 1*time.Minute, stopCh)
```
saikat-royc, line 101: the Run() is already called as a go routing from the main. do we need yet another go routine here?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	shares, err := recon.cloud.File.ListShares(context.TODO(), &file.ListFilter{Project: recon.cloud.Project, Location: "-", InstanceName: "-"})
+	if err != nil {
+		klog.Errorf("Reconciler Failed to list Shares: %v", err)
+	}
+	instances, err := recon.cloud.File.ListMultishareInstances(context.TODO(), &file.ListFilter{Project: recon.cloud.Project, Location: "-"})
+	if err != nil {
+		klog.Errorf("Reconciler Failed to list Instances: %v", err)
+	}
+	klog.V(6).Infof("found %d shares and %d instances", len(shares), len(instances))
+	// ops, err := recon.cloud.File.ListOps(context.TODO(), &file.ListFilter{Project: recon.cloud.Project, Location: "-"})
```
saikat-royc, line 118: remove commented code if not used

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+
+	<-stopCh
+}
+
+func (recon *multishareReconciler) reconcileWorker() {
+	startTime := time.Now()
+
+	shares, err := recon.cloud.File.ListShares(context.TODO(), &file.ListFilter{Project: recon.cloud.Project, Location: "-", InstanceName: "-"})
+	if err != nil {
+		klog.Errorf("Reconciler Failed to list Shares: %v", err)
```
saikat-royc, line 111: should we return if we fail to list shares and instances ?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	var managedInstances []*file.MultishareInstance
+	var managedShares []*file.Share
+
+	instanceShare := make(map[string][]*file.Share)
+	clusterLocation := recon.cloud.Zone
+	if recon.config.IsRegional {
+		var err error
+		clusterLocation, err = util.GetRegionFromZone(clusterLocation)
+		if err != nil {
+			klog.Fatalf("failed to get region for regional cluster: %v", err.Error())
```
saikat-royc, line 758: let's return error from internal functions, and decide on the error handling in the caller

pkg/util/util.go
```diff
@@ -315,3 +315,22 @@ func InstanceStateToCRDStatus(state string) (v1alpha1.FilestoreStatus, error) {
...
 	}
 }
+
+// this function replaces "/" in instanceURI with "." to comply with custom resource naming rule
+func InstanceURIToInstanceInfoName(instanceURI string) string {
+	return strings.ReplaceAll(instanceURI, "/", ".")
+}
+
+// this function replaces "." in instanceInfo name string with "/" to reconstruct instanceURI
+func IINameToInstanceURI(iiName string) string {
```
saikat-royc, line 325: InstanceInfoToInstanceURI

pkg/util/util.go
```diff
@@ -315,3 +315,22 @@ func InstanceStateToCRDStatus(state string) (v1alpha1.FilestoreStatus, error) {
...
+	return strings.ReplaceAll(iiName, ".", "/")
+}
+
+func ShareToSIName(shareName string) string {
+	return strings.ReplaceAll(shareName, "_", "-")
+}
+
+func SIToShareName(siName string) string {
```
saikat-royc, line 333: ShareInfoToShareName

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	return instanceInfoClone, true
+}
+
+func (recon *multishareReconciler) maybeRemoveInstanceInfoFinalizer(instanceInfo *v1alpha1.InstanceInfo) (*v1alpha1.InstanceInfo, error) {
+	if instanceInfo.DeletionTimestamp == nil {
+		klog.V(6).Infof("InstanceInfo %q doesn't have deletion timestamp, it shouldn't be deleted", instanceInfo.Name)
+		return instanceInfo, nil
+	}
+	if instanceInfo.Status == nil {
+		klog.Errorf("InstanceInfo %q marked to be deleted but Status is nil", instanceInfo.Name)
```
saikat-royc, line 604: return error?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		return
+	}
+
+	go wait.Until(recon.reconcileWorker, 1*time.Minute, stopCh)
+
+	<-stopCh
+}
+
+func (recon *multishareReconciler) reconcileWorker() {
+	startTime := time.Now()
```
saikat-royc, line 107: For readability we should comment a summary of the stages

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		klog.Errorf("Filestore CSI driver cannot list InstanceInfo objects: %v", err)
+		return
+	}
+	klog.V(6).Infof("listed out %d instanceInfo objects", len(instanceInfoList))
+	for _, instanceInfo := range instanceInfoList {
+		instanceURI := util.IINameToInstanceURI(instanceInfo.Name)
+		if _, ok := instanceInfoMap[instanceURI]; !ok {
+			instanceInfo, err := recon.maybeRemoveInstanceInfoFinalizer(instanceInfo)
+			if err != nil {
+				klog.Errorf("error removing Finalizer from InstanceInfo %q: %v", instanceInfo.Name, err)
```
saikat-royc, line 152: error removing finalizer ...

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		for _, shareName := range instanceInfo.Status.ShareNames {
+			shareInfo, ok := shareInfos[shareName]
+			if !ok {
+				klog.Errorf("share %q is assigned to instance %q but shareInfo does not exist", shareName, instanceURI)
+				continue
+			}
+			if shareInfo.Status == nil || shareInfo.Status.InstanceHandle == "" {
+				shareInfo, err := recon.assignShareInfo(shareInfo, instanceURI)
+				if err != nil {
+					klog.Errorf("cannot update shareInfo %q's instanceHandle: %v", shareName, err)
```
saikat-royc, line 186: cannot update instanceHandle to shareInfo %q

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+
+		instanceURI := util.IINameToInstanceURI(instanceInfo.Name)
+		for _, shareName := range instanceInfo.Status.ShareNames {
+			shareInfo, ok := shareInfos[shareName]
+			if !ok {
+				klog.Errorf("share %q is assigned to instance %q but shareInfo does not exist", shareName, instanceURI)
+				continue
+			}
+			if shareInfo.Status == nil || shareInfo.Status.InstanceHandle == "" {
+				shareInfo, err := recon.assignShareInfo(shareInfo, instanceURI)
```
saikat-royc, line 184: should we skip this assignment if a share is marked for delete.

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+				if err != nil {
+					klog.Errorf("cannot update shareInfo %q's instanceHandle: %v", shareName, err)
+					continue
+				}
+				shareInfos[shareName] = shareInfo
+			} else if shareInfo.Status.InstanceHandle != instanceURI {
```
saikat-royc, line 190: please put a comment on when can we this potentially happen  `shareInfo.Status.InstanceHandle != instanceURI`

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+					klog.Errorf("failed to update instanceInfo %q: %v", instanceInfo.Name, err)
+					continue
+				}
+				instanceInfos[instanceURI] = instanceInfoClone
+			}
+		}
+	}
+}
+
+func (recon *multishareReconciler) requestInstance(instanceURI string, scTag string) (*v1alpha1.InstanceInfo, error) {
```
saikat-royc, line 327: comment what the function does

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	for _, sc := range storageClasses {
+		klog.V(6).Infof("storageclass %q has Parameters %v", sc.Name, sc.Parameters)
+		if sc.Parameters[driver.ParamMultishareInstanceScLabel] == scTag {
+			return sc, nil
+		}
+	}
+	return nil, fmt.Errorf("no storageclass match storageClassTag %q", scTag)
+}
+
+func (recon *multishareReconciler) assignInstanceInfo(instanceInfo *v1alpha1.InstanceInfo, shareName string) (*v1alpha1.InstanceInfo, error) {
```
saikat-royc, line 365: this function name is confusing. make it more descriptive

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		}
+	}
+
+	recon.assignSharesToInstances(shareInfoMap, instanceInfoMap)
+
+	klog.Infof("reconciliation round finished after %v", time.Since(startTime))
+}
+
+func (recon *multishareReconciler) assignSharesToInstances(shareInfos map[string]*v1alpha1.ShareInfo, instanceInfos map[string]*v1alpha1.InstanceInfo) {
+	for _, instanceInfo := range instanceInfos {
```
saikat-royc, line 167: can we break this function into smaller helper functions.

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+				if err != nil {
+					klog.Errorf("cannot update shareInfo %q's instanceHandle: %v", shareName, err)
+					continue
+				}
+				shareInfos[shareName] = shareInfo
+			} else if shareInfo.Status.InstanceHandle != instanceURI {
+				klog.Warningf("instanceInfo %q has share %q but its shareInfo points to instance %q", instanceInfo.Name, shareName, shareInfo.Status.InstanceHandle)
+
+				// if the share is already marked for deletion, don't try to add it to the instanceInfo it points to.
+				if shareInfo.DeletionTimestamp == nil {
```
saikat-royc, line 194: can we flip the check to avoid indentation?

if shareInfo.Deletetimestamp != nil {
   // remove share from instance info
}

// fix the instanceInfo

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+
+	for instanceURI, instanceInfo := range instanceInfos {
+		if instanceInfo.DeletionTimestamp != nil {
+			continue
+		}
+
+		instanceInfoClone := instanceInfo.DeepCopy()
+		var updated bool
+		if instanceEmpty(instanceInfo) {
+			instanceInfoClone, updated = maybeAddCleanupFinalizer(instanceInfoClone)
```
saikat-royc, line 298: I think we can add the finalizer, during the first creation of the instanceInfo CR itself. It needed not be a separate step?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+			instanceInfo, err = recon.clientset.MultishareV1alpha1().InstanceInfos().Get(context.TODO(), iiName, metav1.GetOptions{})
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					klog.Errorf("error getting instanceInfo %q from api server: %v", iiName, err)
+					continue
+				}
+				klog.V(6).Infof("instanceInfo object for instance %q not found in API server", instance.Name)
+				instanceInfo = nil
+			}
+			klog.V(6).Infof("instanceInfo object for instance %q not found in informer cache but found in API server", instance.Name)
```
saikat-royc, line 419: klog.V(4) ... this may be useful for debugging

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		}
+
+		instanceInfo, err = recon.reconstructInstanceInfo(iiName, instance, instanceInfo)
```
saikat-royc, line 422: reconstructInstanceInfo() is creating the actual CR. Instead, why cant we do a single update to the kube-api server based on the diff of reconstructInstanceInfo() CR and  maybeUpdateInstanceInfoStatus() ?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+				}
+				// shareInfo does not exist for share
+				klog.V(6).Infof("shareInfo object for share %q not found in API server", shareInfoName)
+				shareInfo = nil
+			}
+			// shareInfo exist in api server but not cache
+			klog.V(6).Infof("shareInfo object for share %q not found in informer cache but found in api server", shareInfoName)
+		}
+
+		shareInfo, err = recon.createShareInfo(share, shareInfo)
```
saikat-royc, line 467: why can't we club the createShareInfo and maybeUpdateShareInfoStatus() into one kube-api server update? (i.e generate shareInfo with right status in L467)

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+			continue
+		}
+		instanceInfoMap[instanceURI] = instanceInfo
+	}
+
+	return instanceInfoMap
+}
+
+// ShareInfo objects in the returned map must be treated as read only
+func (recon *multishareReconciler) createAndUpdateShareInfos(shares []*file.Share) map[string]*v1alpha1.ShareInfo {
```
saikat-royc, line 441: Please put a good description of what this function does

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+		newStatus.Error = instanceInfoClone.Status.Error
+	}
+	instanceInfoClone.Status = newStatus
+	klog.Infof("trying to update InstanceInfo %s Status to %v", instanceInfo.Name, instanceInfoClone.Status)
+
+	return recon.updateInstanceInfoStatus(context.TODO(), instanceInfoClone)
+}
+
+func (recon *multishareReconciler) maybeUpdateShareInfoStatus(share *file.Share, shareInfo *v1alpha1.ShareInfo) (*v1alpha1.ShareInfo, error) {
+	status, err := util.ShareStateToCRDStatus(share.State)
```
saikat-royc, line 678: I dont see the definition of ShareStateToCRDStatus()

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+
+func maybeAddCleanupFinalizer(instanceInfoClone *v1alpha1.InstanceInfo) (*v1alpha1.InstanceInfo, bool) {
+	if instanceInfoClone.DeletionTimestamp != nil {
+		klog.Infof("instanceInfo %q has deletionTimestamp so must have already been processed")
+		return instanceInfoClone, false
+	}
+	if len(instanceInfoClone.Finalizers) != 0 {
+		klog.Errorf("instanceInfo %q should not have any finalizer when it's DeletionTimestamp is not set but got %v. Attempting to remove all of them", instanceInfoClone.Name, instanceInfoClone.Finalizers)
+	}
+	instanceInfoClone.Finalizers = []string{util.FilestoreResourceCleanupFinalizer}
```
saikat-royc, line 871: Please elaborate in the definition of FilestoreResourceCleanupFinalizer how this is intended to be used.

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	instanceInfoList, err := recon.instanceLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("Filestore CSI driver cannot list InstanceInfo objects: %v", err)
+		return
+	}
+	klog.V(6).Infof("listed out %d instanceInfo objects", len(instanceInfoList))
+	for _, instanceInfo := range instanceInfoList {
+		instanceURI := util.IINameToInstanceURI(instanceInfo.Name)
+		if _, ok := instanceInfoMap[instanceURI]; !ok {
+			instanceInfo, err := recon.maybeRemoveInstanceInfoFinalizer(instanceInfo)
```
saikat-royc, line 150: Please put some detailed comment, on how the instanceinfo is cleaned up after the finalizer is removed. Its not trivial

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+	}
+
+	for instanceURI, instanceInfo := range instanceInfos {
+		if instanceInfo.DeletionTimestamp != nil {
+			continue
+		}
+
+		instanceInfoClone := instanceInfo.DeepCopy()
+		var updated bool
+		if instanceEmpty(instanceInfo) {
```
saikat-royc, line 297: This is a instance cleanup block. Should this be outside assignSharesToInstances() ?

pkg/util/util.go
leiyiz: done

pkg/util/util.go
leiyiz: done

pkg/multishare_reconciler/reconciler_test.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+			return instanceInfo, nil
+		}
+	}
+	klog.Infof("appending share %q to %q instanceInfo's assigned share list", shareName, instanceInfoClone.Name)
+	instanceInfoClone.Status.ShareNames = append(instanceInfoClone.Status.ShareNames, shareName)
+
+	return recon.updateInstanceInfoStatus(context.TODO(), instanceInfoClone)
+}
+
+func (recon *multishareReconciler) assignShareInfo(shareInfo *v1alpha1.ShareInfo, instanceURI string) (*v1alpha1.ShareInfo, error) {
```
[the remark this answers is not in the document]
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: make sense as it means that we can't update share status anyway.

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: changed the name to `generateInstanceInfo()`. I think since the `createInstanceInfo()` is always called after the instanceInfo is generated, there's no need to split

pkg/multishare_reconciler/reconciler.go
leiyiz: yeah, right now in the most recent code I have this changed to add finalizer at the time of creation. you can see this fix in the coming prs.

pkg/multishare_reconciler/reconciler.go
leiyiz: done

cmd/main.go
leiyiz: in the current code I'm testing I did moved this func into the GCFSDriver. it's a bit hard to port it back to here though

cmd/main.go
leiyiz: the `ensureCustomResourceDefinitionsExist` function needs to have a client passed in, that means the client initiation code needs to also live in main. then main.go it might be a little messy. What do you think?

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: we certainly can move it out of this function but logically we should keep it inside the `assign` funciton.

the reason is that during the previous blocks there're modifications to the instance's share assignment on `instanceInfo.Status.ShareNames` but no resizing were attempted but they are waited to the very end here.

pkg/multishare_reconciler/reconciler.go
leiyiz: note that these blocks must be done in this sequence and won't be reused in general.

done.

pkg/multishare_reconciler/reconciler.go
leiyiz: the "remove share form instanceInfo" step needs to happen either way. So I think flipping the check makes it harder to read.

cmd/main.go
leiyiz: fixed in most recent code version. will be included in upcoming PR

pkg/multishare_reconciler/reconciler.go
leiyiz: yes. done.

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,913 @@
...
+	return recon
+}
+
+func (recon *multishareReconciler) Run(stopCh <-chan struct{}) {
+	klog.Infof("Starting multishare reconciler")
+	defer klog.Infof("Shutting down multishare reconciler")
+
+	informerSynced := []cache.InformerSynced{recon.shareListerSynced, recon.instanceListerSynced}
+	if !cache.WaitForCacheSync(stopCh, informerSynced...) {
+		klog.Errorf("Cannot sync caches")
```
saikat-royc, line 97: Should we fatal here? 
If cache sync fails, and it fails silently and returns (thereby does not start the reconciler worker). Now the CSI calls would stall, because there is no reconciler to react to the CRDs.

cmd/main.go
```diff
@@ -149,3 +171,101 @@ func main() {
...
+		os.Exit(1)
+	}
+	fsClient, err := clientset.NewForConfig(config)
+	if err != nil {
+		klog.Error(err.Error())
+		os.Exit(1)
+	}
+
+	factory := fsInformers.NewSharedInformerFactory(fsClient, resyncPeriod)
+	coreFactory := informers.NewSharedInformerFactory(kubeClient, resyncPeriod)
```
saikat-royc, line 194: let's keep the core factory in main function. It will be needed for other features like 100 shares

cmd/main.go
saikat-royc: Ack let's keep it as is. FYI the kubeclient would be needed for 100shares to create PV informer cache. so let's keep that in main

cmd/main.go
```diff
@@ -58,11 +72,15 @@ var (
...
 	descOverrideMinShareSizeGB = flag.String("desc-override-min-shares-size-gb", "", "If non-empty, the filestore instance description override is used to configure min share size. This flag is ignored if 'feature-max-shares-per-instance' flag is false. Both 'desc-override-max-shares-per-instance' and 'desc-override-min-shares-size-gb' must be provided. 'ecfsDescription' is ignored, if this flag is provided.")
+	enableStateful             = flag.Bool("enable-stateful-multishare", false, "")
+	kubeconfig                 = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Required only when running out of cluster.")
+
 	// This is set at compile time
 	version = "unknown"
 )
 
 const driverName = "filestore.csi.storage.gke.io"
+const resyncPeriod = 15 * time.Minute
```
saikat-royc, line 83: create a flag

cmd/main.go
saikat-royc: hard to port back here you mean in this PR?

pkg/multishare_reconciler/reconciler.go
leiyiz: I wonder if the driver should crash here? although it might send the whole component into crashLooping.

logging worker start is definitely a great point.

cmd/main.go
leiyiz: changed in the most recent commit, will come in upcoming PR

pkg/multishare_reconciler/reconciler.go
saikat-royc: Ack

pkg/multishare_reconciler/reconciler.go
saikat-royc: A log review needs to be done. But if we are clear on logs verbosity which would be useful, let's not introduce more tech debt.

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,873 @@
...
+				if !errors.IsNotFound(err) {
+					klog.Errorf("error getting shareInfo %q from api server: %v", shareInfoName, err)
+					continue
+				}
+				// shareInfo does not exist for share
+				klog.V(6).Infof("shareInfo object for share %q not found in API server", shareInfoName)
+				shareInfo = nil
+			}
+			// shareInfo exist in api server but not cache
+			klog.V(6).Infof("shareInfo object for share %q not found in informer cache but found in api server", shareInfoName)
```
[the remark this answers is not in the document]
saikat-royc: My thought process is logs which help provide insights in prod workflow, and which would not inundate the log file, can be V(4). I am ok with doing a separate log review, but it would be a big task in itself, so lets add logs at V(4) now which we think are useful anyway

pkg/multishare_reconciler/reconciler.go
saikat-royc: yeah let's not crash the entire driver. In L95 we should log something like "starting cache sync ..." also.

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,913 @@
...
+	}
+	klog.V(6).Infof("listed out %d shareInfo objects", len(shareInfoList))
+	for _, shareInfo := range shareInfoList {
+		shareName := shareInfo.Name
+		if _, ok := shareInfoMap[shareName]; !ok {
+			shareInfoMap[shareName] = shareInfo
+		}
+	}
+
+	// create instanceInfo objects if DNE, update instanceInfo.Status based on listed out instances' status
```
saikat-royc, line 143: if does not exist ...

pkg/multishare_reconciler/reconciler.go
saikat-royc: Ack

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,913 @@
...
+}
+
+// assignSharesToEligibleOrNewInstances assigns shares that are not already assigned to eligible instances.
+// If there're no eligible instances, generate a new one.
+func (recon *multishareReconciler) assignSharesToEligibleOrNewInstances(shareInfos map[string]*v1alpha1.ShareInfo, instanceInfos map[string]*v1alpha1.InstanceInfo) {
+	for _, shareInfo := range shareInfos {
+		if shareInfo.Status == nil || shareInfo.Status.InstanceHandle == "" {
+
+			// if share is up to delete, don't assign it even if it's not yet assigned.
+			// This situation should not happen because DeleteVolume shouldn't be called until CreateVolume already succeeded
```
saikat-royc, line 258: We need to write a test case for this. invoke delete for CSI create completes

cmd/main.go
leiyiz: done in most recent code https://github.com/leiyiz/gcp-filestore-csi-driver/blob/stateful_csi/pkg/csi_driver/gcfs_driver.go

pkg/multishare_reconciler/reconciler.go
leiyiz: done.

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: before CSI create completes right? a guitar test might work, I think the main safeguard is that CSI provisioner will make sure that `deleteVolume` is called after `createVolume` return success.

saikat-royc: correct. a test case would ensure there is no regression

pkg/multishare_reconciler/reconciler.go
saikat-royc: Ack. Any ordering assumption please document correctly

pkg/multishare_reconciler/reconciler.go
saikat-royc: Ack

pkg/multishare_reconciler/reconciler_test.go
```diff
@@ -0,0 +1,313 @@
...
+			name:     "nil status instance",
+			instance: &v1alpha1.InstanceInfo{},
+			expected: false,
+		},
+	}
+
+	for _, test := range cases {
+		empty := instanceEmpty(test.instance)
+		if empty != test.expected {
+			t.Errorf("case %s expects %v but got %v", test.name, test.expected, empty)
```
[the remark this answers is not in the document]
leiyiz: the link 404'd but I'm assuming you mean wants... but got ...?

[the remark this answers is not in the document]
leiyiz: done

pkg/multishare_reconciler/reconciler_test.go
```diff
@@ -0,0 +1,313 @@
...
+				},
+			},
+			expected: true,
+		},
+	}
+
+	for _, test := range cases {
+		instanceInfo, updated := maybeAddCleanupFinalizer(test.instance)
+		if updated != test.expected {
+			t.Errorf("case %s expects %v but got %v", test.name, test.expected, updated)
```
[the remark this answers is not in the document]
leiyiz: wouldn't Fatalf also stops the whole test?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,915 @@
...
+	if err != nil {
+		klog.Errorf("Reconciler Failed to list Shares: %v", err)
+		return
+	}
+	instances, err := recon.cloud.File.ListMultishareInstances(context.TODO(), &file.ListFilter{Project: recon.cloud.Project, Location: "-"})
+	if err != nil {
+		klog.Errorf("Reconciler Failed to list Instances: %v", err)
+		return
+	}
+	klog.V(6).Infof("found %d shares and %d instances", len(shares), len(instances))
```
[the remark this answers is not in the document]
leiyiz: done.

cmd/main.go
```diff
@@ -149,3 +171,101 @@ func main() {
...
+		Steps:    steps,
+	}
+	if err := wait.ExponentialBackoff(backoff, condition); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func buildConfig(kubeconfig string) (*rest.Config, error) {
```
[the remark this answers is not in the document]
leiyiz: I think the FIlestore CSI client would require a much higher qps and burst than the node one, and that's edited directly as part of the config. take a look here: https://github.com/leiyiz/gcp-filestore-csi-driver/blob/stateful_csi/pkg/csi_driver/gcfs_driver.go#L281C2-L287

cmd/main.go
```diff
@@ -149,3 +171,101 @@ func main() {
...
+		go recon.Run(stopCh)
+
+		// ...until SIGINT
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+		<-c
+		close(stopCh)
+	}
+
+	go run(context.TODO())
```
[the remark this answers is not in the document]
tyuchn: ack

pkg/multishare_reconciler/reconciler_test.go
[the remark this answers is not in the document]
tyuchn: It will, I thought you wanted to stop the comparison if instanceInfo is not updated as expected, then adding a "continue" should be enough. NVM if my understanding was wrong.

cmd/main.go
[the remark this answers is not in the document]
tyuchn: sg, thanks for pointing this out.

pkg/multishare_reconciler/reconciler.go
leiyiz: done.

pkg/multishare_reconciler/reconciler_test.go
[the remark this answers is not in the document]
leiyiz: done

cmd/main.go
```diff
@@ -58,11 +72,15 @@ var (
 	featureMaxSharePerInstance = flag.Bool("feature-max-shares-per-instance", false, "If this feature flag is enabled, allows the user to configure max shares packed per Filestore instance")
 	descOverrideMaxShareCount  = flag.String("desc-override-max-shares-per-instance", "", "If non-empty, the filestore instance description override is used to configure max share count per instance. This flag is ignored if 'feature-max-shares-per-instance' flag is false. Both 'desc-override-max-shares-per-instance' and 'desc-override-min-shares-size-gb' must be provided. 'ecfsDescription' is ignored, if this flag is provided.")
 	descOverrideMinShareSizeGB = flag.String("desc-override-min-shares-size-gb", "", "If non-empty, the filestore instance description override is used to configure min share size. This flag is ignored if 'feature-max-shares-per-instance' flag is false. Both 'desc-override-max-shares-per-instance' and 'desc-override-min-shares-size-gb' must be provided. 'ecfsDescription' is ignored, if this flag is provided.")
+	enableStateful             = flag.Bool("enable-stateful-multishare", false, "")
```
saikat-royc, line 75: feature-stateful-multishare

cmd/main.go
```diff
@@ -139,6 +157,10 @@ func main() {
 		FeatureOptions:   featureOptions,
 	}
 
+	if *runController && *enableMultishare && *enableStateful {
+		runMultishareReconciler(config)
```
saikat-royc, line 161: why are we starting the reconciler in every replica? It should be started when the driver replica wins the leader election?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+		},
+		Spec: v1alpha1.InstanceInfoSpec{
+			CapacityBytes: instance.CapacityBytes,
+		},
+	}
+	return recon.createInstanceInfo(context.TODO(), instanceInfo)
+}
+
+func (recon *multishareReconciler) createShareInfo(share *file.Share, shareInfo *v1alpha1.ShareInfo) (*v1alpha1.ShareInfo, error) {
+	if shareInfo != nil {
```
saikat-royc, line 547: this logic (if shareInfo != nil) should reside in the caller of this function to keep the code readable

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+	instanceInfoClone.Status = newStatus
+	klog.Infof("Trying to update InstanceInfo %s Status to %v", instanceInfo.Name, instanceInfoClone.Status)
+
+	return recon.updateInstanceInfoStatus(context.TODO(), instanceInfoClone)
+}
+
+func (recon *multishareReconciler) maybeUpdateShareInfoStatus(share *file.Share, shareInfo *v1alpha1.ShareInfo) (*v1alpha1.ShareInfo, error) {
+	status, err := util.ShareStateToCRDStatus(share.State)
+	if err != nil {
+		return shareInfo, err
```
saikat-royc, line 725: we should return nil, err in L725?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+	shareInfoClone := shareInfo.DeepCopy()
+	if shareInfoClone.Status == nil {
+		shareInfoClone.Status = &v1alpha1.ShareInfoStatus{}
+	}
+	shareInfoClone.Status.InstanceHandle = instanceURI
+	klog.Infof("Try to assign share %q to instance %q", shareInfo.Name, instanceURI)
+	return recon.updateShareInfoStatus(context.TODO(), shareInfoClone)
+}
+
+// createAndUpdateInstanceInfos create instanceInfo objects if needed and updates their statuses to match with reality.
```
saikat-royc, line 432: nit: "... to match with the real state of the world"

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+			klog.Errorf("Error updating instanceInfo %q: %v", iiName, err)
+			continue
+		}
+		instanceInfoMap[instanceURI] = instanceInfo
+	}
+
+	return instanceInfoMap
+}
+
+// createAndUpdateShareInfos create shareInfo objects if needed and updates their statuses to match with reality.
```
saikat-royc, line 481: nit: ... to match with the real state of the world

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+			continue
+		}
+		shareInfoMap[shareInfoName] = shareInfo
+	}
+
+	return shareInfoMap
+}
+
+func (recon *multishareReconciler) reconstructInstanceInfo(iiName string, instance *file.MultishareInstance, instanceInfo *v1alpha1.InstanceInfo) (*v1alpha1.InstanceInfo, error) {
+	if instanceInfo != nil {
```
saikat-royc, line 528: this logic of if instanceInfo != nil should reside in the caller for readability

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+
+	instanceInfo = &v1alpha1.InstanceInfo{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: iiName,
+			Labels: map[string]string{
+				driver.ParamMultishareInstanceScLabel: instance.Labels[util.ParamMultishareInstanceScLabelKey],
+			},
+		},
+		Spec: v1alpha1.InstanceInfoSpec{
+			CapacityBytes: instance.CapacityBytes,
```
saikat-royc, line 540: In generateInstanceInfo() we put the storage class name in spec.StorageClassName, don't we need it here?

pkg/multishare_reconciler/reconciler.go
saikat-royc: I agree that keeping the finalizer is safer. My reasoning is we should not get into this state where the deletetimestamp != nil and status == nil. (because in instanceEmpty() we return false for nil status).  It could be a user invoked delete. Is this reasoning valid?

klog.Warningf() may be a better log level?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+	recon.deleteOrResizeInstances(instanceInfos)
+}
+
+// fixTwoWayPointers scans over all instanceInfo objects and try to fix the 2 way pointer between instanceInfo and shareInfo objects.
+// instanceInfo.Status.ShareNames -> shareInfo
+// shareInfo.Status.InstanceHandle -> instanceInfo
+func (recon *multishareReconciler) fixTwoWayPointers(shareInfos map[string]*v1alpha1.ShareInfo, instanceInfos map[string]*v1alpha1.InstanceInfo) {
+	for _, instanceInfo := range instanceInfos {
+		if instanceInfo.Status == nil {
+			klog.Infof("Instance %q has Status nil", instanceInfo.Name)
```
saikat-royc, line 191: klog.Infof is too verbose here?

pkg/multishare_reconciler/reconciler.go
```diff
@@ -0,0 +1,917 @@
...
+	recon.assignSharesToEligibleOrNewInstances(shareInfos, instanceInfos)
+
+	// Have to call deleteOrResizeInstances() after assigning shares and/or fixing two way pointers because no resizing were attempted in
+	// assignSharesToEligibleOrNewInstances() or fixTwoWayPointers()
+	recon.deleteOrResizeInstances(instanceInfos)
+}
+
+// fixTwoWayPointers scans over all instanceInfo objects and try to fix the 2 way pointer between instanceInfo and shareInfo objects.
+// instanceInfo.Status.ShareNames -> shareInfo
+// shareInfo.Status.InstanceHandle -> instanceInfo
```
saikat-royc, line 187: In the comment we should also mention the 2 way ptr fix prioritizes share parent reference and fixes the instance to point to the right share, and why

pkg/multishare_reconciler/reconciler.go
leiyiz: done

cmd/main.go
leiyiz: fixed in [branch for next PR](https://github.com/leiyiz/gcp-filestore-csi-driver/blob/stateful_reconciler/cmd/main.go#L75)

pkg/multishare_reconciler/reconciler.go
leiyiz: done.

pkg/multishare_reconciler/reconciler.go
leiyiz: I think instance could yet to be created so it can have nil status. changed to V(6) to only log for debugging

pkg/multishare_reconciler/reconciler.go
leiyiz: we don't, because the only need for `StorageClassName` is to get parameter when calling Filestore API to create a new instance. since Instance already exist here, there's no need for StorageClassName here.

pkg/multishare_reconciler/reconciler.go
leiyiz: fixed.

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: done

pkg/multishare_reconciler/reconciler.go
leiyiz: do you mean the user could manually mark the object for deletion?

leiyiz: I think perhaps returning `instanceInfo` as is and `nil` err is better: 1) it keeps instanceInfo as is on the record for reconciler and 2) we already logged the Warning message, no need to log out some error message again

cmd/main.go
leiyiz: added TODO within `runMultishareReconciler` to fix in the next PR

pkg/multishare_reconciler/reconciler.go
saikat-royc: Please address the reason as a comment in the followup PR.

leiyiz: in the reconstruct case, the instance is already present with a `scTag`. And we cannot change the instances' property anymore. we don't need to have `storageClassName` field (the storageclass might not exist either) and it will act like a distinguisher between migrated instances and driver-created instances.

PR #447, opened 2023-03-08
Fix backup source comparison logic for single share instances

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

If the user uses static provisioning to create filestore instance with a share name which is != "vol1", and the csi driver hits a code path where the backup already exists (say timeout during creation of backup and retry by the csi-snapshotter sidecar), the csi driver uses a logic to compare source volume id handle with a hard coded "vol1". if the driver gets into this state, the volume snapshot object will never converge to true.

Fixes #

```release-note
Fix backup source comparison logic for single share instances
```

### DIFF
--- a/pkg/cloud_provider/file/fake.go
+++ b/pkg/cloud_provider/file/fake.go
@@ -164,8 +164,9 @@ func (manager *fakeServiceManager) CreateBackup(ctx context.Context, obj *Servic
 
 	backupSource := fmt.Sprintf("projects/%s/locations/%s/instances/%s", obj.Project, obj.Location, obj.Name)
 	if backupInfo, ok := manager.backups[backupUri]; ok {
-		if backupInfo.SourceVolumeHandle != backupSource {
-			return nil, fmt.Errorf("Mismatch in source volume handle for existing snapshot")
+		if backupInfo.SourceInstance != backupSource && backupInfo.SourceShare != obj.Volume.Name {
+			// TODO: format the right info
+			return nil, fmt.Errorf("Mismatch in source for existing snapshot %v", backupInfo)
 		}
 		return backupInfo.Backup, nil
 	}
@@ -179,8 +180,9 @@ func (manager *fakeServiceManager) CreateBackup(ctx context.Context, obj *Servic
 		CapacityGb:      defaultCapacityGb,
 	}
 	manager.backups[backupUri] = &BackupInfo{
-		Backup:             backupToCreate,
-		SourceVolumeHandle: backupSource,
+		Backup:         backupToCreate,
+		SourceInstance: backupSource,
+		SourceShare:    obj.Volume.Name,
 	}
 	return backupToCreate, nil
 }
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -862,12 +862,12 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 	} else {
-		backupSourceCSIHandle, err := util.BackupVolumeSourceToCSIVolumeHandle(backupInfo.SourceVolumeHandle)
+		backupSourceCSIHandle, err := util.BackupVolumeSourceToCSIVolumeHandle(backupInfo.SourceInstance, backupInfo.SourceShare)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "Cannot determine volume handle from back source %s", backupInfo.SourceVolumeHandle)
+			return nil, status.Errorf(codes.Internal, "Cannot determine volume handle from back source instance %s, share %s", backupInfo.SourceInstance, backupInfo.SourceShare)
 		}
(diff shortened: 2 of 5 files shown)


pkg/cloud_provider/file/fake.go
```diff
@@ -164,8 +164,9 @@ func (manager *fakeServiceManager) CreateBackup(ctx context.Context, obj *Servic
 
 	backupSource := fmt.Sprintf("projects/%s/locations/%s/instances/%s", obj.Project, obj.Location, obj.Name)
 	if backupInfo, ok := manager.backups[backupUri]; ok {
-		if backupInfo.SourceVolumeHandle != backupSource {
-			return nil, fmt.Errorf("Mismatch in source volume handle for existing snapshot")
+		if backupInfo.SourceInstance != backupSource && backupInfo.SourceShare != obj.Volume.Name {
```
leiyiz, line 167: `obj.Volume.Name` only works for single share right?

saikat-royc: Yes. We have not yet implemented multishare backups. That would have a different function

pkg/csi_driver/controller.go
```diff
@@ -862,12 +862,12 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
...
 		}
 	} else {
-		backupSourceCSIHandle, err := util.BackupVolumeSourceToCSIVolumeHandle(backupInfo.SourceVolumeHandle)
+		backupSourceCSIHandle, err := util.BackupVolumeSourceToCSIVolumeHandle(backupInfo.SourceInstance, backupInfo.SourceShare)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "Cannot determine volume handle from back source %s", backupInfo.SourceVolumeHandle)
+			return nil, status.Errorf(codes.Internal, "Cannot determine volume handle from back source instance %s, share %s", backupInfo.SourceInstance, backupInfo.SourceShare)
 		}
 		if backupSourceCSIHandle != volumeID {
-			return nil, status.Errorf(codes.AlreadyExists, "Backup already exists with a different source volume %s, input source volume %s", backupInfo.SourceVolumeHandle, volumeID)
```
hsadoyan, line 870: Why do we choose to log the csiHandle here instead of the Source Volume Handle?

pkg/csi_driver/controller_test.go
```diff
@@ -1048,22 +1048,100 @@ func ValidateExpectedError(t *testing.T, errResp <-chan error, operationUnblocke
...
 func TestCreateSnapshot(t *testing.T) {
+	type BackupInfo struct {
+		s              *file.ServiceInstance
+		backupName     string
+		backupLocation string
+	}
+	backupName := "mybackup"
+	project := "test-project"
+	zone := "us-central1-c"
+	_ = zone
```
hsadoyan, line 1059: What does the `_ = zone` do? Isn't the underscore usually for unused variables?

saikat-royc: good catch

pkg/csi_driver/controller.go
saikat-royc: So that we understand what exactly was compared in L869
also I removed backupInfo.SourceVolumeHandle

saikat-royc: Because we are comparing against `backupSourceCSIHandle` in L869.
I removed the backupInfo.SourceVolumeHandle field.

Is that your question?

hsadoyan: Makes sense. So the log should already have been logging `backupSourceCSIHandle` instead of `backupInfo.SourceVolumeHandle`

I just wasn't sure if it's directly related to this change, or just a drive-by cleanup.

pkg/csi_driver/controller_test.go
```diff
@@ -1048,22 +1048,99 @@ func ValidateExpectedError(t *testing.T, errResp <-chan error, operationUnblocke
...
 		{
 			name: "Create snapshot request for multishare backed volumes",
 			req: &csi.CreateSnapshotRequest{
-				SourceVolumeId: "modeMultishare/mysc/myproject/location/instancename/sharename",
+				SourceVolumeId: "modeMultishare/mysc/test-project/location/myinstance/myshare",
 			},
 			expectErr: true,
 		},
+		{
+			name: "Existing backup found, with same different volume Id (source zonal filestore instance), error expected",
```
hsadoyan, line 1077:
```suggestion
			name: "Existing backup found, with different volume Id (source zonal filestore instance), error expected",
```

saikat-royc: Fixed

pkg/csi_driver/controller.go
saikat-royc: Ack. you are correct, logs should have already logging the variable thats being compared in the if logic

leiyiz (commented):
/lgtm
/approve

PR #461, opened 2023-03-24
Implement configurable max shares CreateVolume path with feature gate

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Second PR in a series of PR to implement the configurable max shares/instance feature. In this PR, the CreateVolume code path is implemented and gated under the feature flag.
Key changes:
1. validate parse the storage class config, and generate the max supported shares. 
2. Plumb the max shares to the instance create call
3. Implement the description override flag to setup maxshares, min share size knobs (which is needed until filestore is ready with instance.MaxSharesCount field)
4. If feature is disabled, then maxShareCount is set to 0. This translates to a default value of 10 shares/instance in filestore.

Fixes #

Future PRs will include
1. ControllerExpandVolume [changes](https://docs.google.com/document/d/1Jsx2H90NPjP8kddrljwwi0LKIlAmkT_ImAa_f4mwUfI/edit#bookmark=id.o8s7hdo9or5a)
2. Poll Timeout [changes](https://docs.google.com/document/d/1Jsx2H90NPjP8kddrljwwi0LKIlAmkT_ImAa_f4mwUfI/edit#bookmark=id.f59tfd9c6m8)

For this PR, 
1. tested manually that feature disabled path works as expected. Instance maxShareCount returns 10.
2. Tested description override. Instance maxShareCount returns value according to the description override knob.

Yet to test instance.maxShareCount during create. Needs an allowlist for that.

```release-note
Implement configurable max shares CreateVolume path with feature gate
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -71,6 +71,7 @@ type MultishareInstance struct {
 	State              string
 	KmsKeyName         string
 	Description        string
+	MaxShareCount      int
 }
 
 func (i *MultishareInstance) String() string {
@@ -803,10 +804,11 @@ func (manager *gcfsServiceManager) StartCreateMultishareInstanceOp(ctx context.C
 				ConnectMode:     instance.Network.ConnectMode,
 			},
 		},
-		CapacityGb:  util.BytesToGb(instance.CapacityBytes),
-		KmsKeyName:  instance.KmsKeyName,
-		Labels:      instance.Labels,
-		Description: instance.Description,
+		CapacityGb:    util.BytesToGb(instance.CapacityBytes),
+		KmsKeyName:    instance.KmsKeyName,
+		Labels:        instance.Labels,
+		Description:   instance.Description,
+		MaxShareCount: int64(instance.MaxShareCount),
 	}
 
 	op, err := manager.multishareInstancesService.Create(locationURI(instance.Project, instance.Location), targetinstance).InstanceId(instance.Name).Context(ctx).Do()
@@ -1088,6 +1090,7 @@ func cloudInstanceToMultishareInstance(instance *filev1beta1multishare.Instance)
 		MaxCapacityBytes:   instance.MaxCapacityGb * util.Gb,
 		CapacityStepSizeGb: instance.CapacityStepSizeGb,
 		Description:        instance.Description,
+		MaxShareCount:      int(instance.MaxShareCount),
 	}, nil
 }
 
--- a/pkg/csi_driver/multishare_controller.go
+++ b/pkg/csi_driver/multishare_controller.go
@@ -18,12 +18,14 @@ package driver
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/klog/v2"
 	cloud "sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider"
@@ -39,6 +41,9 @@ const (
 	methodExpandVolume              = "ExpandVolume"
 	ecfsDataPlaneVersionFormat      = "GoogleReserved-CustomVMImage=clh.image.ems.path:projects/%s/global/images/ems-filestore-scaleout-%s"
 	ecfsCustom100sharesConfigFormat = "GoogleReservedOverrides={\"CustomMultiShareConfig\":{\"MaxShareCount\": %d, \"MinShareSizeGB\":%d}}"
+
+	// volume context attributes
+	attrMaxShareSize = "max-share-size"
 )
 
 // MultishareController handles CSI calls for volumes which use Filestore multishare instances.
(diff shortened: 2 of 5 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -1088,6 +1090,7 @@ func cloudInstanceToMultishareInstance(instance *filev1beta1multishare.Instance)
 		MaxCapacityBytes:   instance.MaxCapacityGb * util.Gb,
 		CapacityStepSizeGb: instance.CapacityStepSizeGb,
 		Description:        instance.Description,
+		MaxShareCount:      int(instance.MaxShareCount),
```
leiyiz, line 1093: for this, do we want to add logic to have a default value (10?) if the `instance.MaxShareCount` is not populated?

pkg/csi_driver/multishare_controller.go
```diff
@@ -440,14 +470,19 @@ func (m *MultishareController) generateNewMultishareInstance(instanceName string
 		KmsKeyName:  kmsKeyName,
 		Labels:      labels,
 		Description: generateInstanceDescFromEcfsDesc(m.ecfsDescription),
-	}, nil
+	}
+	if m.featureMaxSharePerInstance {
```
leiyiz, line 474: would it be better to condition on maxShareCount?
if maxShareCount == 0 ....

that way there's a easy way to specify "just use the default value"?

pkg/csi_driver/multishare_controller.go
```diff
@@ -440,14 +470,19 @@ func (m *MultishareController) generateNewMultishareInstance(instanceName string
...
+	return f, nil
 }
 
 func generateNewShare(name string, parent *file.MultishareInstance, req *csi.CreateVolumeRequest) (*file.Share, error) {
 	if parent == nil {
 		return nil, status.Error(codes.Internal, "parent mulishare instance is empty")
 	}
-	targetSizeBytes, err := getShareRequestCapacity(req.CapacityRange)
+	// The share size request is already validated in CreateVolume call
+	targetSizeBytes, err := getShareRequestCapacity(req.CapacityRange, util.MinShareSizeBytes, util.MaxShareSizeBytes)
```
leiyiz, line 485: don't we need an if clause here to pass in `util.MinShareSizeConfigurableBytes` and user-specified max range here?

pkg/csi_driver/multishare_controller.go
```diff
@@ -590,6 +625,9 @@ func (m *MultishareController) generateCSICreateVolumeResponse(instancePrefix st
 	if m.driver.config.FeatureOptions.FeatureLockRelease.Enabled {
 		resp.Volume.VolumeContext[attrSupportLockRelease] = "true"
 	}
+	if m.featureMaxSharePerInstance {
+		resp.Volume.VolumeContext[attrMaxShareSize] = strconv.Itoa(int(maxShareSizeSizeBytes))
```
leiyiz, line 629: what is this used for? i assume in the future when `ControllerExpandVolume` is called this context will be used to determine if it's pass limit?

pkg/csi_driver/multishare_controller_test.go
```diff
@@ -1621,3 +1800,214 @@ func TestMultishareControllerExpandVolume(t *testing.T) {
...
 	}
 }
+
+func TestIsValidMaxVolSize(t *testing.T) {
+	tests := []struct {
+		val         int64
+		expectedret bool
+	}{
+		{
+			val: 0,
```
leiyiz, line 1810: consider add case for non-expected unit like `1024Gb` or `131072Mb`

consider add case for unsupported stop like `384 Gb`

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -224,6 +224,10 @@ func (m *MultishareOpsManager) startInstanceWorkflow(ctx context.Context, w *Wor
 	}
 	switch w.opType {
 	case util.InstanceCreate:
+		if m.msControllerServer != nil && !m.msControllerServer.featureMaxSharePerInstance {
+			// MaxShareCount field set to 0, defaults to 10 shares in Filestore
+			w.instance.MaxShareCount = 0
```
leiyiz, line 229: this seem to be extra step since MaxShareCount is already set to 0 before?

pkg/csi_driver/multishare_controller.go
saikat-royc: I want to avoid indirect inference of the feature.

pkg/csi_driver/multishare_controller.go
saikat-royc: yes it will be in the next PR

saikat-royc: as mentioned in the comment, by the time we reach to the `generateNewShare` function we have already validated the range. So we do not need a strict range change. Hence I have kept the function as it is.

pkg/csi_driver/multishare_controller.go
saikat-royc: Correct. ExpandVolume does not have the volume context, so we will infer it from the PV spec.csi.volumecontext

pkg/csi_driver/multishare_ops_manager.go
saikat-royc: this is an extra safeguard to set the value to 0 if feature disabled. thoughts?

pkg/csi_driver/multishare_controller.go
leiyiz: what I mean is that if the feature is turned on, and user ask for 10Gi, here the `util.MinShareSizeBytes` passed into `getShareRequestCapacity` would mean the 10Gi is below the min size and produced error right?

pkg/csi_driver/multishare_controller.go
```diff
@@ -263,7 +290,8 @@ func (m *MultishareController) ControllerExpandVolume(ctx context.Context, req *
 		return nil, status.Error(codes.InvalidArgument, "ControllerExpandVolume volume ID must be provided")
 	}
 
-	reqBytes, err := getShareRequestCapacity(req.GetCapacityRange())
+	// TODO: This logic will change to derive the max allowable share size from the PV.spec.CSI.volumeAttributes[“max-share-size”]
+	reqBytes, err := getShareRequestCapacity(req.GetCapacityRange(), util.MinShareSizeBytes, util.MaxShareSizeBytes)
```
leiyiz, line 294: same here

pkg/cloud_provider/file/file.go
saikat-royc: that would again need an if check for feature flag. hence kept that logic in the controller

pkg/csi_driver/multishare_ops_manager.go
saikat-royc: Ack. The generateNewMultishareInstance checks for the feature flag before setting the MaxShareCount. Removing this extra check here

pkg/csi_driver/multishare_controller_test.go
saikat-royc: Good point

pkg/csi_driver/multishare_controller.go
saikat-royc: Fixed it

PR #462, opened 2023-03-29
Enable parsing regional backup location

hsadoyan:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Enables Enterprise tier backups and adds unit tests for creating volumes from a snapshot

Fixes #

```release-note
Enable parsing regional backup locations and enable Enterprise tier backups
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -514,7 +514,7 @@ func (manager *gcfsServiceManager) GetBackup(ctx context.Context, backupUri stri
 }
 
 func (manager *gcfsServiceManager) CreateBackup(ctx context.Context, obj *ServiceInstance, backupName string, backupLocation string) (*filev1beta1.Backup, error) {
-	backupUri, region, err := CreateBackpURI(obj, backupName, backupLocation)
+	backupUri, region, err := CreateBackupURI(obj, backupName, backupLocation)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -695,16 +695,37 @@ func PollOpErrorCode(err error) *codes.Code {
...
+	region, err := deduceRegion(obj, backupLocation)
+	if err != nil {
+		return "", "", err
+	}
+	return backupURI(obj.Project, region, backupName), region, nil
+}
+
+// deduceRegion will either return the provided backupLocation region or deduce
+// from the object
+func deduceRegion(obj *ServiceInstance, backupLocation string) (string, error) {
```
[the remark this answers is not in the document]
hsadoyan: Added

pkg/csi_driver/controller_test.go
```diff
@@ -1150,6 +1381,87 @@ func TestCreateSnapshot(t *testing.T) {
...
+		name           string
+		backupName     string
+		backupLocation string
+		instance       *file.ServiceInstance
+		expectedURL    string
+		expectedRegion string
+		expectErr      bool
+	}{
+		{
+			name:           "Region is not provided. ServiceInstance is regional.",
```
saikat-royc, line 1400: Some negative test cases as well needed here

hsadoyan: Good call. 

Actually, thinking about it, I have a question. Are we worried about cases where the provided `backupLocation` param is not a valid region? Right now if one is provided we just take it blindly as `region`. 

And if we are worried about that case, and a zone is provided instead of region there, do we want to crash out or take the region from the zone.

hsadoyan: Since this is a user provided field, my instinct is to error here, instead of trying to be cute, so I'll make that change. Open to other opinions though

hsadoyan: Pushed.

saikat-royc: If the location is unexpected, error is the right behavior

PR #478, opened 2023-04-12
stateful CSI call logic

leiyiz:
**What type of PR is this?**
/kind feature

1) csi call logic for Stateful driver
2) changed instanceInfo reconciling and share assignment logic in the case that user manually deleted instance
3) added filtering logic for error message older than 15 min

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/multishare_stateful_controller.go
+++ b/pkg/csi_driver/multishare_stateful_controller.go
@@ -1,0 +1,300 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	csi "github.com/container-storage-interface/spec/lib/go/csi"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/apis/multishare/v1alpha1"
+	clientset "sigs.k8s.io/gcp-filestore-csi-driver/pkg/client/clientset/versioned"
+	listers "sigs.k8s.io/gcp-filestore-csi-driver/pkg/client/listers/multishare/v1alpha1"
+	cloud "sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/file"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
+)
+
+// MultishareController handles CSI calls for volumes which use Filestore multishare instances.
+type MultishareStatefulController struct {
+	driver *GCFSDriver
+	zone   string
+	cloud  *cloud.Cloud
+	mc     *MultishareController
+
+	clientset   clientset.Interface
+	shareLister listers.ShareInfoLister
+}
+
+func NewMultishareStatefulController(config *controllerServerConfig) *MultishareStatefulController {
+	return &MultishareStatefulController{
+		driver:      config.driver,
+		zone:        config.cloud.Zone,
+		cloud:       config.cloud,
+		clientset:   config.features.FeatureStateful.DriverClientSet,
+		shareLister: config.features.FeatureStateful.ShareLister,
+	}
+}
+
+func (m *MultishareStatefulController) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
+	klog.Infof("CreateVolume called for multishare with request %+v", req)
+	pvName := req.GetName()
+	if len(pvName) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "CreateVolume name must be provided")
+	}
+	if err := m.driver.validateVolumeCapabilities(req.GetVolumeCapabilities()); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if req.GetVolumeContentSource() != nil {
+		return nil, status.Error(codes.InvalidArgument, "Multishare backed volumes do not support volume content source")
+	}
+
+	instanceSCLabel, err := getInstanceSCLabel(req)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	reqBytes, err := getShareRequestCapacity(req.GetCapacityRange(), util.MinShareSizeBytes, util.MaxShareSizeBytes)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if !util.IsAligned(reqBytes, util.Gb) {
+		return nil, status.Errorf(codes.InvalidArgument, "requested size(bytes) %d is not a multiple of 1GiB", reqBytes)
+	}
+
+	shareInfo, err := m.shareLister.Get(pvName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, status.Errorf(codes.Internal, "error getting shareInfo %q from informer: %s", pvName, err.Error())
+		}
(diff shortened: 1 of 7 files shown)


pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/apis/multishare/v1alpha1"
+	clientset "sigs.k8s.io/gcp-filestore-csi-driver/pkg/client/clientset/versioned"
+	listers "sigs.k8s.io/gcp-filestore-csi-driver/pkg/client/listers/multishare/v1alpha1"
+	cloud "sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/file"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
+)
+
```
saikat-royc, line 34: put a todo here to incorporate 80 shares / instance support

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	if !util.IsAligned(reqBytes, util.Gb) {
+		return nil, status.Errorf(codes.InvalidArgument, "requested size(bytes) %d is not a multiple of 1GiB", reqBytes)
+	}
+
+	shareInfo, err := m.shareLister.Get(pvName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, status.Errorf(codes.Internal, "error getting shareInfo %q from informer: %s", pvName, err.Error())
+		}
```
saikat-royc, line 86: Add a log here that we are querying api server

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if !util.IsAligned(reqBytes, util.Gb) {
+		return nil, status.Errorf(codes.InvalidArgument, "requested size(bytes) %d is not a multiple of 1GiB", reqBytes)
+	}
+
+	shareInfo, err := m.shareLister.Get(pvName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, status.Errorf(codes.Internal, "error getting shareInfo %q from informer: %s", pvName, err.Error())
```
saikat-royc, line 85: shareInfo for volume %q

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	klog.V(4).Infof("DeleteVolume called for multishare with request %+v", req)
+
+	siName := util.ShareToShareInfoName(shareName)
+	shareInfo, err := m.shareLister.Get(siName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, status.Errorf(codes.Internal, "error deleting volume %q due to informer error: %s", req.VolumeId, err.Error())
+		}
+		// check with api if share exist
+		klog.V(6).Infof("shareInfo %s DNE in cache, checking if share is already deleted", siName)
```
saikat-royc, line 165: does not exist

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		_, err := m.cloud.File.GetShare(ctx, &file.Share{
+			Parent: &file.MultishareInstance{
+				Project:  project,
+				Location: location,
+				Name:     instanceName,
+			},
+			Name: shareName,
+		})
+		if err != nil {
+			// If share not found, proceed to instance/shrink check.
```
saikat-royc, line 175: remove comment

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		return nil, status.Errorf(codes.Aborted, "waiting to express intent for volume %s to be deleted", req.VolumeId)
+	}
+
+	if shareInfo.DeletionTimestamp == nil {
+		if len(shareInfo.Finalizers) == 0 {
+			klog.Errorf("shareInfo %s shouldn't have no finalizer before deletion marking", siName)
+			return nil, status.Errorf(codes.Internal, "error deleting volume %s due to driver state error", req.VolumeId)
+		}
+		// delete shareInfo
```
saikat-royc, line 190: remove comment

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		err := m.deleteShareInfo(ctx, siName)
+		if err != nil {
+			klog.Errorf("error marking the shareInfo object as deleted: %s", err.Error())
+		}
+		return nil, status.Errorf(codes.Aborted, "expressed intent for volume %s to be deleted, waiting.", req.VolumeId)
+	}
+
+	if shareInfo.Status == nil {
+		klog.Errorf("shareInfo %s marked to be deleted but shareInfo.Status == nil", siName)
+		return nil, status.Errorf(codes.Aborted, "waiting for volume %s to be deleteed.", req.VolumeId)
```
saikat-royc, line 200: typo: "deleted"

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+			return nil, status.Errorf(codes.Internal, "error deleting volume %s due to failed internal update", req.VolumeId)
+		}
+		return &csi.DeleteVolumeResponse{}, nil
+	}
+
+	if shareInfo.Status.Error != "" {
+		return nil, status.Errorf(codes.Internal, "internal error: %s", shareInfo.Status.Error)
+	}
+
+	return nil, status.Errorf(codes.Aborted, "waiting for share %s to be deleted", req.VolumeId)
```
saikat-royc, line 220: be clear about the resource name.

if we are using share then we should use shareInfo.spec.sharename

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, status.Errorf(codes.Aborted, "expressed intent for volume %s to be expanded", siName)
+	}
+
+	if shareInfo.Status == nil {
+		klog.Errorf("ControllerExpandVolume called for %s but shareInfo.Status is nil", siName)
+		return nil, status.Errorf(codes.Internal, "volume %s is not yet created", siName)
+	}
+
+	if shareInfo.Status.CapacityBytes >= reqBytes && shareInfo.Status.ShareStatus == v1alpha1.READY {
+		klog.Infof("Controller expand volume succeeded for volume %v, size(bytes): %v", req.VolumeId, shareInfo.Status.CapacityBytes)
```
saikat-royc, line 285: we should also verify with getshare() the real share capacity just to be sure (https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/multishare_controller.go#L352)

pkg/csi_driver/reconciler.go
```diff
@@ -1240,8 +1231,17 @@ func (recon *MultishareReconciler) listMultishareResourceOps(ctx context.Context
...
 
+		klog.V(6).Infof("creation time: %s", meta.CreateTime)
 		var err error
 		if op.Error != nil {
+			// filter out error Op that's more than util.ErrRetention old
+			var createTime time.Time
+			createTime, err = time.Parse(util.OpTimeLayout, meta.CreateTime)
+			if err != nil {
+				klog.Errorf("failed to parse creation Time %q with error: %s", meta.CreateTime, err.Error())
+			} else if createTime.Before(time.Now().Add(-util.ErrRetention)) {
```
saikat-royc, line 1242: 1. why are we skipping ops before some threshold time?
2. how was 15 min decided?

leiyiz: we are only skipping `if op.Error != nil`. 

For now it's arbitrary and doesn't do much. The thinking is that once the error is old, it's less relevant and we should reconcile the crds' `status.Error` field to prevent having error status on this shareInfo or instanceInfo forever and pushing up error counts.

But for now I haven't included the logic to actually remove the `status.Error`

saikat-royc: just to be safe we can also check (op.Done && op.Error != nil) in L1236

pkg/csi_driver/reconciler.go
```diff
@@ -1240,8 +1231,17 @@ func (recon *MultishareReconciler) listMultishareResourceOps(ctx context.Context
 			continue
```
saikat-royc, line 1231: if we fail to parse metadata we may risk skipping running ops?

leiyiz: yeah but then we wouldn't be able to tell the target of the Op anyway: `file.IsInstanceTarget(meta.Target)`

saikat-royc: ok

PR #483, opened 2023-04-20
LockInfo CRD for NFS lock release

tyuchn:
**What type of PR is this?**
>
kind feature

LockInfo CRD definition and the corresponding client.

Fixes #

N/A

```release-note
This PR introduces a CRD which will be used to store mounted filestore info for each node.
```


pkg/apis/lockrelease/register.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package lockrelease
+
+// GroupName is the group name used in this package
+const (
+	GroupName = "filestore.csi.storage.gke.io"
```
leiyiz, line 20: IIRC the group name needs to be unique, using `filestore.csi.storage.gke.io` might result in conflicts in the future, consider using something like `lock.filestore.csi.storage.gke.io`

PR #521, opened 2023-05-22
Replace PollOpErrorCode and IsUserError with CodeForError

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Replace PollOpErrorCode and IsUserError with CodeForError, and filter out User errors from ControllerExpandVolume.This makes things more concise, and closely models what is done in the PDCSI driver. This will be followed up by a PR that filters out the rest of the context / user errors from the multishare controller. 

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -30,6 +30,7 @@ import (
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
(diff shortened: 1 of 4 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -681,15 +681,22 @@ func IsContextError(err error) *codes.Code {
...
+// (2) "context canceled", returns grpc Canceled
+func CodeForError(err error) *codes.Code {
+	if errCode := isUserError(err); errCode != nil {
 		return errCode
 	}
-	if errCode := IsContextError(err); errCode != nil {
+	if errCode := isContextError(err); errCode != nil {
 		return errCode
 	}
 	return util.ErrCodePtr(codes.Internal)
```
saikat-royc, line 702: Can we also check if there is already some known status.code set in the err and skip overwrite of code to Internal.

if s, ok := status.FromError(test.err); ok {
 ... if s.Code() == // one of [codes](https://github.com/grpc/grpc-go/blob/v1.55.0/codes/codes.go)
}

amacaskill: Done!

judemars:
/cherry-pick release-1.4

(use this one not the one above, I got confused and deleted that one)

PR #523, opened 2023-05-23
Use CodeForError when we call filestore API

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Use CodeForError when we call filestore API. Additionally, use CodeForError to surface error codes returned from helper functions.  This will ensure that if the helper functions are changed in the future to return different error codes, those error codes will be surfaced. I also attempted to change some of the internal error codes to invalid argument, but will get feedback on whether that was correct. 

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -625,8 +625,8 @@ func GetInstanceNameFromURI(uri string) (project, location, name string, err err
 }
 
 func IsNotFoundErr(err error) bool {
-	apiErr, ok := err.(*googleapi.Error)
-	if !ok {
+	var apiErr *googleapi.Error
+	if !errors.As(err, &apiErr) {
 		return false
 	}
 
@@ -718,7 +718,7 @@ func existingErrorCode(err error) *codes.Code {
 	return nil
 }
 
-// CodeForError returns a pointer to the grpc error code that maps to the http
+// codeForError returns a pointer to the grpc error code that maps to the http
 // error code for the passed in user googleapi error or context error. Returns
 // codes.Internal if the given error is not a googleapi error caused by the user.
 // The following http error codes are considered user errors:
@@ -729,7 +729,10 @@ func existingErrorCode(err error) *codes.Code {
 // The following errors are considered context errors:
 // (1) "context deadline exceeded", returns grpc DeadlineExceeded,
 // (2) "context canceled", returns grpc Canceled
-func CodeForError(err error) *codes.Code {
+func codeForError(err error) *codes.Code {
+	if err == nil {
+		return nil
+	}
 	if errCode := existingErrorCode(err); errCode != nil {
 		return errCode
 	}
@@ -741,6 +744,15 @@ func CodeForError(err error) *codes.Code {
 	}
 
 	return util.ErrCodePtr(codes.Internal)
+}
+
+// Status error returns the error as a grpc status error, and
+// sets the grpc error code according to CodeForError.
+func StatusError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return status.Error(*codeForError(err), err.Error())
 }
 
 // This function returns the backup URI, the region that was picked to be the backup resource location and error.
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -214,7 +214,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 			_, err = s.config.fileService.GetBackup(ctx, id)
 			if err != nil {
 				klog.Errorf("Failed to get volume %v source snapshot %v: %v", name, id, err.Error())
-				return nil, status.Error(*file.CodeForError(err), err.Error())
+				return nil, file.StatusError(err)
 			}
 			sourceSnapshotId = id
 		}
@@ -224,7 +224,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 	filer, err := s.config.fileService.GetInstance(ctx, newFiler)
 	// No error is returned if the instance is not found during CreateVolume.
 	if err != nil && !file.IsNotFoundErr(err) {
-		return nil, status.Error(*file.CodeForError(err), err.Error())
+		return nil, file.StatusError(err)
 	}
 
 	if filer != nil {
@@ -263,7 +263,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 			// In case of abort, the CIDR IP is released and available for reservation
 			defer s.config.ipAllocator.ReleaseIPRange(reservedIPRange)
 			if err != nil {
-				return nil, err
+				return nil, file.StatusError(err)
 			}
 
 			// Adding the reserved IP range to the instance object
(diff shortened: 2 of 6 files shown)


pkg/csi_driver/controller.go
```diff
@@ -894,7 +894,7 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
 		}
 		tp, err := util.ParseTimestamp(backupInfo.Backup.CreateTime)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to parse create timestamp for backup %v", backupInfo.Backup.Name)
+			return nil, status.Errorf(*file.CodeForError(err), "failed to parse create timestamp for backup %v", backupInfo.Backup.Name)
```
amacaskill, line 897: We are returning status error codes from within util.ParseTimestamp. If we ever change/add a different error code to  `ParseTimestamp` in the future, we want it to be surfaced.

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -648,7 +649,7 @@ func containsOpWithShareName(shareName string, opType util.OperationType, ops []
 func containsOpWithShareTarget(share *file.Share, opType util.OperationType, ops []*OpInfo) (*OpInfo, error) {
 	shareUri, err := file.GenerateShareURI(share)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to parse share handle, err: %v", err.Error())
```
amacaskill, line 651: @saikat-royc Should all of the errors returned from GenerateShareURI really be an internal error? Or can "missing share parent instance" or "missing parent, project, name or location in share parent" be InvalidArgument?

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -251,7 +251,7 @@ func (m *MultishareOpsManager) startInstanceWorkflow(ctx context.Context, w *Wor
 func (m *MultishareOpsManager) verifyNoRunningInstanceOps(instance *file.MultishareInstance, ops []*OpInfo) error {
 	instanceUri, err := file.GenerateMultishareInstanceURI(instance)
```
amacaskill, line 252: @saikat-royc Should all of the errors returned from GenerateMultishareInstanceURI be internal errors? Or can some be InvalidArgument?

pkg/cloud_provider/file/file.go
```diff
@@ -720,6 +723,15 @@ func CodeForError(err error) *codes.Code {
 	return util.ErrCodePtr(codes.Internal)
 }
 
+// Status error returns the error as a grpc status error, and
+// sets the grpc error code according to CodeForError.
+func StatusError(err error) error {
```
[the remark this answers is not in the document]
saikat-royc: StatusError seems cleaner

pkg/csi_driver/multishare_ops_manager.go
```diff
@@ -215,7 +215,7 @@ func (m *MultishareOpsManager) startInstanceWorkflow(ctx context.Context, w *Wor
 	// 1. verify no instance ops or share (belonging to the instance) ops running for the given instance.
 	// 2. Start the instance op.
 	if w.instance == nil {
-		return nil, status.Errorf(codes.Internal, "instance not found in workflow object")
+		return nil, status.Errorf(codes.InvalidArgument, "instance not found in workflow object")
```
saikat-royc, line 218: this would be internalError (ihandled by StatusForError()) since we do not expect the instance object to be nil here?

pkg/csi_driver/controller.go
```diff
@@ -262,7 +262,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 			// In case of abort, the CIDR IP is released and available for reservation
 			defer s.config.ipAllocator.ReleaseIPRange(reservedIPRange)
 			if err != nil {
-				return nil, err
+				return nil, status.Error(*file.CodeForError(err), err.Error())
```
saikat-royc, line 265: So all of these can be return nil, StatusError(err)

pkg/csi_driver/controller.go
```diff
@@ -917,7 +920,7 @@ func (s *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSn
 	}
```
saikat-royc, line 920: L915-L917

Multiple nested ifs
```
	if err != nil {
		if err != nil {
		}
	}
```

pkg/csi_driver/multishare_controller.go
```diff
@@ -303,7 +315,7 @@ func (m *MultishareController) startAndWaitForInstanceDeleteOrShrink(ctx context
 	}
```
saikat-royc, line 315: startAndWaitForInstanceDeleteOrShrink() calls some ops manager functions which need the StatusError() handling

pkg/csi_driver/multishare_controller.go
```diff
@@ -513,7 +533,7 @@ func (m *MultishareController) generateNewMultishareInstance(instanceName string
 
 func generateNewShare(name string, parent *file.MultishareInstance, req *csi.CreateVolumeRequest) (*file.Share, error) {
 	if parent == nil {
-		return nil, status.Error(codes.Internal, "parent mulishare instance is empty")
+		return nil, status.Error(codes.InvalidArgument, "parent multishare instance is empty")
```
saikat-royc, line 536: Internal is the right code

pkg/csi_driver/multishare_controller.go
```diff
@@ -644,7 +664,7 @@ func getShareRequestCapacity(capRange *csi.CapacityRange, minShareSizeBytes, max
 func (m *MultishareController) generateCSICreateVolumeResponse(instancePrefix string, s *file.Share, maxShareSizeBytes int64) (*csi.CreateVolumeResponse, error) {
 	volId, err := generateMultishareVolumeIdFromShare(instancePrefix, s)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, err.Error())
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
```
saikat-royc, line 667: Internal is the right code

pkg/csi_driver/multishare_controller.go
amacaskill: I don't think any functions in the ops manager need the StatusError handling, if we wrap errors appropriately, then we can just call StatusError in multishare_controller only.

pkg/csi_driver/multishare_controller.go
```diff
@@ -210,15 +212,16 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
...
-	return m.getShareAndGenerateCSICreateVolumeResponse(ctx, instanceScPrefix, newShare, maxShareSizeSizeBytes)
+	resp, err := m.getShareAndGenerateCSICreateVolumeResponse(ctx, instanceScPrefix, newShare, maxShareSizeSizeBytes)
+	return resp, file.StatusError(err)
 }
 
 func (m *MultishareController) getShareAndGenerateCSICreateVolumeResponse(ctx context.Context, instancePrefix string, s *file.Share, maxShareSizeSizeBytes int64) (*csi.CreateVolumeResponse, error) {
 	share, err := m.cloud.File.GetShare(ctx, s)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, err.Error())
+		return nil, err
```
amacaskill, line 224: @saikat-royc Is this correct?

pkg/csi_driver/multishare_controller.go
saikat-royc: Ack. This is cleaner to handle it only in controller

PR #524, opened 2023-05-23
Fix lock release logs

tyuchn:
**What type of PR is this?**
>
/kind bug

Fix lock release logs

Fixes #

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -528,21 +528,21 @@ func (s *nodeServer) nodeStageVolumeUpdateLockInfo(ctx context.Context, req *csi
 	filestoreIP := attr[attrIP]
 	if cm == nil {
 		data := map[string]string{lockInfoKey: filestoreIP}
-		klog.Infof("NodeStageVolume creating configmap %s/%s with data %v for volume %s", util.ManagedFilestoreCSINamespace, configmapName, data, volumeID)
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/node.go
```diff
@@ -542,7 +542,7 @@ func (s *nodeServer) nodeStageVolumeUpdateLockInfo(ctx context.Context, req *csi
 	}
 
 	if err := s.lockReleaseController.UpdateConfigMapWithKeyValue(ctx, cm, lockInfoKey, filestoreIP); err != nil {
-		klog.Errorf("NodeStageVolume failed to update configmap %s/%s with lock info {%s: %s} for volume %s: %v", util.ManagedFilestoreCSINamespace, configmapName, volumeID, err)
+		klog.Errorf("NodeStageVolume failed to update configmap %s/%s with lock info {%s: %s} for volume %s: %v", util.ManagedFilestoreCSINamespace, configmapName, lockInfoKey, filestoreIP, volumeID, err)
```
msau42, line 545: You can use klog.KObj to simplify the name/namespace: https://github.com/kubernetes/klog/blob/6ded8085b27eeaf68a13d583a382b125ef1482ee/k8s_references.go#LL78C7-L78C7

tyuchn: Done. Thanks for the recommendation.

PR #526, opened 2023-05-30
Allow to add extra volume labels from command line option

tsmetana:
**What type of PR is this?**
/kind feature

This is basically a clone of the feature of GCP PD added with the https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/693.

The patch adds a new command line option `extra-labels` allowing users to automatically add custom labels to the disks provisioned by the driver. Main motivation is to be able to clearly denote cloud resource ownership and allow for automated resource management.

```release-note
A new `extra-labels` command line option has been added. It can be used to attach extra labels to the provisioned volumes.
```


jsafrane:
@saikat-royc, can you please take a look?

tsmetana:
The test keeps failing with `SSH encountered an error: Instance gcfs-csi-e2e-us-central1-c in state RUNNING but not available by SSH` and subsequent `failed to extract test archive` errors. I won't attempt to /retest since this looks like it's not a temporary issue (something broke in the CI?). It doesn't seem to have anything to do with the proposed code changes though.

leiyiz (commented):
/approve

PR #534, opened 2023-06-06
Return InvalidArgument code for NodeStageVolume error caused by VPC network unalignment

tyuchn:
**What type of PR is this?**
kind cleanup

NodeStageVolume usually fails because of "Filestore instance not in the same VPC network with the GKE cluster". This one exists in 90% of unhappy cluster, which frequently trigger our SLO alert. Instead of aggressively assuming `status 32 == VPC error`, we'll just pop an error message in this PR.
**Which issue(s) this PR fixes**:`

Fixes #

```release-note
N/A
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -313,7 +313,7 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 	if needsCreateDir {
 		klog.V(4).Infof("NodeStageVolume attempting mkdir for path %s", stagingTargetPath)
 		if err := os.MkdirAll(stagingTargetPath, 0750); err != nil {
-			return nil, fmt.Errorf("mkdir failed for path %s (%w)", stagingTargetPath, err)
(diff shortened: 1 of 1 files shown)


pkg/csi_driver/node.go
```diff
@@ -331,6 +331,11 @@ func (s *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 		if unmntErr := mount.CleanupMountPoint(stagingTargetPath, s.mounter, false /* extensiveMountPointCheck */); unmntErr != nil {
 			klog.Errorf("Unmount %q failed: %v", stagingTargetPath, unmntErr.Error())
 		}
+		isVPCErr := strings.Contains(err.Error(), "exit status 32")
```
saikat-royc, line 334: 32 may not be just for vpc errors

tyuchn: Agreed, I should not aggressively make such assumption. So I'll simply pop a message to check VPC for now. Hopefully this can be solved once we improve static provisioned volumes as our H2 plan.

saikat-royc: Ack

tyuchn:
Close since this PR is no longer required

PR #563, opened 2023-07-12
high scale tier ip reservation support and remove cmek support check

leiyiz:
**What type of PR is this?**
/kind feature
code change required to support high scale tier

Fixes #

`Warning  ProvisioningFailed    3s (x5 over 19s)      filestore.csi.storage.gke.io_gke-test-recon-default-pool-9053d3d5-11rf_0a08c303-a5d9-49c5-834e-8a3867a44617  failed to provision volume with StorageClass "csi-filestore": rpc error: code = InvalidArgument desc = googleapi: Error 400: tier BASIC_HDD does not support KMS encryption., badRequest`

```release-note
CMEK support now won't be checked in the CSI driver, trying to create basic or premium tier instances with cmek will result in invalid argument error from the Filestore API.

high scale tier instance creation with IP reservation is now supported
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -309,6 +309,9 @@ func (s *controllerServer) reserveIPRange(ctx context.Context, filer *file.Servi
 	if filer.Tier == enterpriseTier {
 		ipRangeSize = util.IpRangeSizeEnterprise
 	}
+	if filer.Tier == highScaleTier {
+		ipRangeSize = util.IpRangeSizeHighScale
+	}
 	unreservedIPBlock, err := s.config.ipAllocator.GetUnreservedIPRange(cidr, ipRangeSize, cloudInstancesReservedIPRanges)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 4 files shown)


pkg/csi_driver/controller.go
```diff
@@ -601,9 +604,6 @@ func (s *controllerServer) generateNewFileInstance(name string, capBytes int64,
 			return nil, fmt.Errorf("invalid parameter %q", k)
 		}
 	}
-	if kmsKeyName != "" && tier != enterpriseTier {
```
saikat-royc, line 604: why not add a (tier != enterpriseTier && tier != HighScaleTier) ?

leiyiz: this way we can avoid future potential misalignments with the Filestore API. for example if Filestore change what tier has cmek capability we don't have to change our code anymore.

saikat-royc: Ack. But if user tries out KMS with non compatible tiers, we would create SLO noise

saikat-royc: [isUserError()](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/cloud_provider/file/file.go#L649) should ideally handle it. 

Please verify manually if it fails for a incompatible tier and capture the logs in this PR for future reference

leiyiz: done

PR #580, opened 2023-08-10
Multishare Create Backup

hsadoyan:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This allows us to create multishare backups 

Fixes #

Main Changes:

- Rename the current `BackupInfo` struct to `Backup`.
- Introduce new `BackupInfo` struct that contains all of the information needed to create a new backup.
- Refactor Singleshare CreateBackup to use the new `backupinfo` struct instead of overloading the ServiceInstance. 
- Refactor backup related helper methods to no longer take a `ServiceInstance` object. 
- Add Multishare CreateBackup
- Add labels to multishare createbackup
- Update unit tests to no longer use service instance. The `initialBackup` field was being used incorrectly/unintuitively. Clarify what each test is testing.

```release-note
Implement creating multishare backup
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
@@ -109,10 +110,38 @@ type Network struct {
 	Ip              string
 }
 
-type BackupInfo struct {
+type Backup struct {
 	Backup         *filev1beta1.Backup
 	SourceInstance string
 	SourceShare    string
+}
+
+type BackupInfo struct {
+	Name               string
+	SourceVolumeId     string
+	BackupURI          string
+	SourceInstance     string
+	SourceInstanceName string
+	SourceShare        string
+	Project            string
+	Location           string
+	Tier               string
+	Labels             map[string]string
+}
+
+func (bi *BackupInfo) SourceVolumeLocation() string {
+	splitId := strings.Split(bi.SourceVolumeId, "/")
+	// Format: "modeInstance/us-central1/myinstance/myshare",
+
+	return splitId[1]
+}
+
+func (bi *BackupInfo) BackupSource() string {
+	if isMultishareVolId(bi.SourceVolumeId) {
+		return shareURI(bi.Project, bi.SourceVolumeLocation(), bi.SourceInstanceName, bi.SourceShare)
+	} else {
+		return instanceURI(bi.Project, bi.SourceVolumeLocation(), bi.SourceInstanceName)
+	}
 }
 
 type Service interface {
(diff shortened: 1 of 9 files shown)


pkg/csi_driver/controller_test.go
```diff
@@ -1252,76 +1269,87 @@ func ValidateExpectedError(t *testing.T, errResp <-chan error, operationUnblocke
...
+				SourceVolumeId: "modeInstance/us-central1/myinstance/myshare",
 				Name:           backupName,
 				Parameters: map[string]string{
 					util.VolumeSnapshotTypeKey: "backup",
 				},
 			},
-			initialBackup: &BackupInfo{
-				s: &file.ServiceInstance{
-					Project:  project,
-					Location: zone,
```
hsadoyan, line 1292: While the initial service instance can be a zone, the initialBackup itself cannot be. Having the service instance in the backupInfo made the tests more difficult to parse through.

pkg/util/util.go
```diff
@@ -50,7 +50,8 @@ const (
 	snapshotTotalElements = 6
 
 	// number of elements in backup Volume sources e.g. projects/{project name}/locations/{zone}/instances/{name}
-	volumeTotalElements = 6
+	singleShareVolumeTotalElements = 6
+	multiShareVolumeTotalElements  = 8
```
leiyiz, line 54: can you add comment here to give an example of the string

pkg/cloud_provider/file/file.go
```diff
@@ -504,55 +533,59 @@ func (manager *gcfsServiceManager) ResizeInstance(ctx context.Context, obj *Serv
...
 }
 
+func (manager *gcfsServiceManager) CreateMultishareBackup(ctx context.Context, backupInfo *BackupInfo) (*filev1beta1.Backup, error) {
+
+	return manager.CreateBackup(ctx, backupInfo)
+}
+
+func (manager *gcfsServiceManager) CreateSingleshareBackup(ctx context.Context, backupInfo *BackupInfo) (*filev1beta1.Backup, error) {
+	return manager.CreateBackup(ctx, backupInfo)
+}
```
leiyiz, line 587: why do we need separate functions when the content are the same?

hsadoyan: Oh, I forgot to delete the bodies here. The content was different before refactor, and they aren't exported anymore. Good catch

pkg/util/util.go
hsadoyan: Added.

pkg/csi_driver/multishare_controller.go
```diff
@@ -218,6 +220,101 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
...
+
+		snapshot, err := file.ProcessExistingBackup(ctx, existingBackup, volumeID, modeMultishare)
+		if err != nil {
+			return nil, err
+		}
+		return &csi.CreateSnapshotResponse{
+			Snapshot: snapshot,
+		}, nil
+	} else {
+		//no existing Snapshot
```
leiyiz, line 268: nit: maybe "no existing backup"?

pkg/csi_driver/multishare_controller.go
```diff
@@ -218,6 +220,101 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
...
+		}
+		return resp, nil
+	}
+}
+
+func (m *MultishareController) createNewBackup(ctx context.Context, backupInfo *file.BackupInfo) (*csi.Snapshot, error) {
+
+	backupObj, err := m.cloud.File.CreateBackup(ctx, backupInfo)
+	if err != nil {
+		klog.Errorf("Create snapshot for volume Id %s failed: %v", backupInfo.SourceVolumeId, err.Error())
```
leiyiz, line 300: What would happen if the backup creation took some time and timed out? what is the error code returned here? is it abort or internal or something else? concern is that backup creation could push up the SLO signal

hsadoyan: That's a good question, and I'm not sure. We didn't have any special case handling for singleshare backups, so I didn't put any here. I'll investigate.

hsadoyan: Discussed with the FS team. They aren't expecting timeouts from this, since the SLO is on the order of hours/days depending on size of volume.

leiyiz (commented):
/approve

PR #586, opened 2023-08-18
zonal large tier support

leiyiz:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
now supports tier "zonal" with large band. small band is actively blocked
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -42,6 +42,7 @@ const (
 	basicHDDTier   = "basic_hdd"
 	basicSSDTier   = "basic_ssd"
 	highScaleTier  = "high_scale_ssd"
+	zonalTier      = "zonal"
 	defaultNetwork = "default"
 
 	defaultTierMinSize    = 1 * util.Tb
@@ -50,6 +51,8 @@ const (
 	enterpriseTierMaxSize = 10 * util.Tb
 	highScaleTierMinSize  = 10 * util.Tb
 	highScaleTierMaxSize  = 100 * util.Tb
+	zonalTierMinSize      = 1 * util.Tb
+	zonalTierMaxSize      = 100 * util.Tb
 	premiumTierMinSize    = 25 * util.Tb / 10
 	premiumTierMaxSize    = 639 * util.Tb / 10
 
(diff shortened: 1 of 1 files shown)


pkg/csi_driver/controller.go
```diff
@@ -50,6 +51,8 @@ const (
 	enterpriseTierMaxSize = 10 * util.Tb
 	highScaleTierMinSize  = 10 * util.Tb
```
saikat-royc, line 52: Ideally we do not want to support both "high_scale_ssd" and "zonal" string. (i.e if user provides "high_scale_ssd" in the storage class we should deny the request). However let's double check with Filestore on what is the end of life for this stirng. The deprecation path for this would be tricky, where filestore would deprecate first and then GKE would rollout a change to not support the "high_scale_ssd" string in storageclass, and needs auto-upgrades etc. So we want to avoid all of that and just support the "zonal" string.
 
As long as static provisioning is not impacted by this, we should only support 1 tier string.

We need to map it to the definition of the bands https://cloud.google.com/filestore/docs/service-tiers#zonal_with_capacity_bands

leiyiz: We have customers using `high_scale_ssd` now and we were not actively blocking it, blocking now would be a breaking change

pkg/csi_driver/controller.go
```diff
@@ -50,6 +51,8 @@ const (
...
 	highScaleTierMaxSize  = 100 * util.Tb
```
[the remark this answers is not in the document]
leiyiz: I don't see why we should have 4 different bounds - we can use the existing `highScaleTierMinSize` to serve as the switching point

pkg/csi_driver/controller.go
saikat-royc: Yes, thats the concern. Does tier impact volume expansion or delete calls? 

Asking because, say starting GKE 1.x we support tier "zonal" only, it impacts only new PVC creation. Existing ones continue to work throughout its lifecycle.

leiyiz: offline followed up, `high_scale_ssd` won't be deprecated.

pkg/csi_driver/controller.go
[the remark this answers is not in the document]
saikat-royc: Ack. zonalHighCapacityMaxSize, zonalHighCapacityMinSize is sufficient

Lets not use "highScaleTier*" names and align with the filestore names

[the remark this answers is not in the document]
leiyiz: Followed up offline, for future backwards compatibility, we want to keep `highScaleTierMinSize` and `highScaleTierMaxSize`.

For now the switchpoint can reuse `highScaleTierMinSize` and we can add `zonalHighCapacityMinSize` if we think it's getting complex

PR #596, opened 2023-08-29
Enable or Enhance Dependabot

zhuxiaow0:
Enable Dependabot for Docker and GoMod if not already so.
Keeping K8s dependencies up-to-date. 

Signed-off-by: zhuxiaow@google.com

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: docker
+  directory: /
+  schedule:
+    interval: daily
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: daily


.github/dependabot.yaml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: docker
+  directory: /
+  schedule:
+    interval: daily
```
sunnylovestiramisu, line 6: Do we want to update it daily? This will create PR and email spam for reviewers.

Please consider using the beta feature for grouping: https://github.com/kubernetes-csi/external-provisioner/blob/master/.github/dependabot.yaml#L2-L22

And also we should set a limit of the max PR it creates via [open-pull-requests-limit](https://github.com/kubernetes-csi/external-provisioner/blob/master/.github/dependabot.yaml#L36C3-L36C31)

PR #642, opened 2023-10-12
Enable dependabot

tyuchn:
**What type of PR is this?**
>
/kind feature

The dependabot will have a weekly scan of the following files on master branch:
- go.mod
- dockerfile
- dockerfile for webhook

Fixes #

```release-note
N/A
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,54 @@
+version: 2
+enable-beta-ecosystems: true
+updates:
+- package-ecosystem: "docker"
+  directory: "/"
+  schedule:
+    interval: weekly
+  labels:
+    - "area/dependency"
+    - "release-note-none"
+    - "ok-to-test"
+    - "kind/cleanup"
+- package-ecosystem: "docker"
+  directory: "/cmd/webhook"
+  schedule:
+    interval: weekly
+  labels:
+    - "area/dependency"
+    - "release-note-none"
+    - "ok-to-test"
+    - "kind/cleanup"
+- package-ecosystem: gomod
+  directory: "/"
+  allow:
+  - dependency-type: "all"
+  schedule:
+    interval: weekly
+  groups:
+    github-google:
+      patterns:
+        - "github.com/google*"
+    golang-x:
+      patterns:
+        - "golang.org/x*"
+    google-golang:
+      patterns:
+        - "google.golang.org*"
+    onsi:
+      patterns:
+        - "github.com/onsi*"
+    spf13:
+      patterns:
+        - "github.com/spf13*"
+    k8s-dependencies:
+      patterns:
+        - "k8s.io*"
+        - "sigs.k8s.io*"
+        - "github.com/kubernetes-csi*"
+  labels:
+    - "area/dependency"
+    - "release-note-none"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+  directory: "/"
+  schedule:
+    interval: weekly
+  labels:
+    - "area/dependency"
+    - "release-note-none"
+    - "ok-to-test"
+    - "kind/cleanup"
+- package-ecosystem: "docker"
+  directory: "/cmd/webhook"
```
sunnylovestiramisu, line 14: Is this only "/cmd/webhook" or "/cmd/webhook*"?

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+        - "github.com/golang*"
+    k8s-dependencies:
+      patterns:
+        - "k8s.io*"
+        - "sigs.k8s.io*"
+        - "github.com/kubernetes-csi*"
+  labels:
+    - "area/dependency"
+    - "release-note-none"
+    - "ok-to-test"
```
sunnylovestiramisu, line 40: Do you also want to group github.com: https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/go.mod#L7-L16 or it does not matter?

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+    interval: weekly
+  groups:
+    golang-dependencies:
+      patterns:
+        - "github.com/golang*"
+    k8s-dependencies:
+      patterns:
+        - "k8s.io*"
+        - "sigs.k8s.io*"
+        - "github.com/kubernetes-csi*"
```
sunnylovestiramisu, line 36: You can remove "github.com/kubernetes-csi*" since filestore go.mod does not have dependency on that path.

sunnylovestiramisu: Oh ignore this one, it is actually there a few lines above.

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
+version: 2
+enable-beta-ecosystems: true
+updates:
+- package-ecosystem: "docker"
```
sunnylovestiramisu, line 4: You need to add open-pull-requests-limit for each package-ecosystem

.github/dependabot.yml
tyuchn: `cmd/webhook` should be fine here for Dockerfile scanning.

.github/dependabot.yml
tyuchn: Done. I assume that at most one PR will be opened for `package-ecosystem: docker``, because the go version is the only thing that needs to be updated. But I agree there's no harm to set a limitation here.

.github/dependabot.yml
tyuchn: Grouping all packages under github.com seems dangerous because the functionalities of the packages vary too much. However, it is a good point to group packages with similar functionalities in the same group, such as `github.com/onsi` and `github.com/spf13`. I will group some of the packages with better granularity. Please take another look at the modified groups. We can experiment and modify them on the fly.

PR #758, opened 2023-12-22
[release-1.3] Update go modules

sunnylovestiramisu:
Updated the following go modules:
golang.org/x/crypto@v0.17.0
```release-note
NONE
```


msau42:
crypto module doesn't seem to be used in this branch

sunnylovestiramisu:
/close

the golang.org/x/crypto is not used here

PR #774, opened 2024-01-08
Disable NFSExportOptions by default

hsadoyan:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Disable the IP ACL by default until we get launch approval

Fixes #

```release-note
Disable IP ACL configuration by default
```

### DIFF
--- a/pkg/csi_driver/multishare_controller.go
+++ b/pkg/csi_driver/multishare_controller.go
@@ -57,16 +57,17 @@ const (
 
 // MultishareController handles CSI calls for volumes which use Filestore multishare instances.
 type MultishareController struct {
-	driver                     *GCFSDriver
-	fileService                file.Service
-	cloud                      *cloud.Cloud
-	opsManager                 *MultishareOpsManager
-	volumeLocks                *util.VolumeLocks
-	ecfsDescription            string
-	isRegional                 bool
-	clustername                string
-	featureMaxSharePerInstance bool
-	featureMultishareBackups   bool
+	driver                          *GCFSDriver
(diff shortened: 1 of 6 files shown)


pkg/csi_driver/multishare_controller.go
```diff
@@ -57,16 +57,17 @@ const (
...
+	fileService                     file.Service
+	cloud                           *cloud.Cloud
+	opsManager                      *MultishareOpsManager
+	volumeLocks                     *util.VolumeLocks
+	ecfsDescription                 string
+	isRegional                      bool
+	clustername                     string
+	featureMaxSharePerInstance      bool
+	featureMultishareBackups        bool
+	FeatureNFSExportOptionsOnCreate bool
```
leiyiz, line 70: not sure why `featureMaxSharePerInstance` and `featureMultishareBackups` are lower cases here but consider align with them and use lower case

leiyiz (commented):
/approve

pkg/csi_driver/multishare_controller.go
hsadoyan: Good catch. Fixed

leiyiz: I don't see it reflected for some reason, can you double check

hsadoyan: Oops. Now actually fixed

leiyiz (commented):
/lgtm

PR #864, opened 2024-05-31
Properly unwrap gce-compute error code.

hime:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

A new way of wrapping errors was introduced in [compute engine API](https://cloud.google.com/compute/docs/reference/rest/v1) (eg. "cloud.service.Disks.Insert") where some of these errors are wrapped with an additional layer of googleapi.Error. This additional wrap does not contain the http error code in its error code field, but instead contains a default value of Unknown. This is problematic because one of the ways we extract errors is by checking if the error can be a status (signifying a temporary error instead). The way the error is structured allows it to be classified as temporary, which causes our code to return the Unknown error present in the outermost layer, before any other error checks occur.

This has been causing issues in how we report errors coming from compute engine, since we do not currently unwrap them properly, resulting in Unknown error code being reported in most scenarios. This PR ensures we are unwrapping the embedded error code properly while making sure all the other error types we encounter continue to process errors as expected.

Fixes #

```release-note
Error codes extracted from errors part of compute engine api are now exposed with correct http errors instead of being classified as Unknown.
```

### DIFF
--- a/pkg/cloud_provider/file/file.go
+++ b/pkg/cloud_provider/file/file.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/googleapis/gax-go/v2/apierror"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
@@ -214,6 +215,14 @@ var (
 	instanceUriRegex = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/instances/([^/]+)$`)
 	shareUriRegex    = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/instances/([^/]+)/shares/([^/]+)$`)
 )
+
+// userErrorCodeMap tells how API error types are translated to error codes.
+var userErrorCodeMap = map[int]codes.Code{
+	http.StatusForbidden:       codes.PermissionDenied,
+	http.StatusBadRequest:      codes.InvalidArgument,
+	http.StatusTooManyRequests: codes.ResourceExhausted,
+	http.StatusNotFound:        codes.NotFound,
+}
 
 func NewGCFSService(version string, client *http.Client, primaryFilestoreServiceEndpoint, testFilestoreServiceEndpoint string) (Service, error) {
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


pkg/cloud_provider/file/file.go
```diff
@@ -215,6 +216,14 @@ var (
 	shareUriRegex    = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/instances/([^/]+)/shares/([^/]+)$`)
 )
 
+// userErrorCodeMap tells how API error types are translated to error codes.
+var userErrorCodeMap = map[int]codes.Code{
```
amacaskill, line 220: Can you please change isUserError() to use this map too?

amacaskill: I think maybe we could change isUserError() to isUserOperationError() as these occur when errors happen when polling the operations (https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/535) . I think we only need isUserOperationError() to include this logic: 

```
	var apiErr *googleapi.Error
	if !errors.As(err, &apiErr) {
		// Fallback to check for expected error code in the error string
		return containsUserErrStr(err)
	}
```

A little background if you were curious... we already have this [logic integrated into waitForOp in the PDCSI Driver](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/bf3736fb82427da66a6126e4deecfdaeee1e300e/pkg/gce-cloud-provider/compute/gce-compute.go#L1127), so that's why we don't need to do this operation logic in PDCSI CodeForError. PDCSI's implementation is nicer, because we could easily convert the op error to a computev1.OperationErrorErrors, but IIRC, Filestore did some weird error wrapping for operations, so the best way to fix this was to string match.

pkg/cloud_provider/file/file_test.go
```diff
@@ -536,11 +537,52 @@ func TestIsContextError(t *testing.T) {
 }
 
 func TestCodeForError(t *testing.T) {
```
amacaskill, line 539: It might also be nice to add 1 testcase for isUserOperationError into CodeForError, so we can make sure codeForError works all together. Right now we are testing each function within codeForError individually, but we should also add one test case for each function so that we know it works e2e.

hime: Added the non-overlapping test cases from isUserOperationError into CodeForError test.

pkg/cloud_provider/file/file.go
hime: I updated the PR to remove any overlap, let me know what you think!

PR #924, opened 2024-08-07
Change error reporting to temporary in CreateVolume workflow

### LINKED ISSUE #925: Change error reporting to Temporary while Creating Volume
When `CreateVolume` is called, the `csi-provisioner` expects either "final" or "temporary" errors. If we report final error when Filestore instance creation could be ongoing or it is already created, then a PVC deletion before successful volume creation could lead to resource leakage as DeleteVolume won't happen.

This is already done for [gcp-compute-persistent-disk-csi-driver](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver). 

PRs: 

- https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1558
- https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1664

riteshghorse:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When `CreateVolume` is called, the `csi-provisioner` expects either "final" or "temporary" errors. If we report final error when Filestore instance creation could be ongoing or it is already created, then a PVC deletion before successful volume creation could lead to resource leakage as DeleteVolume won't happen.

Fixes #925

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -241,7 +241,9 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 	filer, err := s.config.fileService.GetInstance(ctx, newFiler)
 	// No error is returned if the instance is not found during CreateVolume.
 	if err != nil && !file.IsNotFoundErr(err) {
-		return nil, file.StatusError(err)
+		// Failed to GetInstance, however the Filestore instance may already be created.
+		// The error should be non-final.
+		return nil, status.Errorf(codes.Unavailable, "CreateVolume, failed to get instance: %v", err.Error())
 	}
 
 	if filer != nil {
@@ -259,11 +261,11 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		if filer.State != "READY" {
 			msg := fmt.Sprintf("Volume %v not ready, current state: %v", name, filer.State)
 			klog.V(4).Infof(msg)
-			return nil, status.Error(codes.Internal, msg)
+			return nil, status.Error(codes.Unavailable, msg)
 		}
 	} else {
 		param := req.GetParameters()
-		// If we are creating a new instance, we need pick an unused CIDR range from reserved-ipv4-cidr
+		// If we are creating a new instance, we need to pick an unused CIDR range from reserved-ipv4-cidr
 		// If the param was not provided, we default reservedIPRange to "" and cloud provider takes care of the allocation
 		if newFiler.Network.ConnectMode == privateServiceAccess {
 			if reservedIPRange, ok := param[ParamReservedIPRange]; ok {
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/klog/v2"
 	cloud "sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/file"
+	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/common"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/metrics"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/util"
 )
@@ -243,7 +244,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 	if err != nil && !file.IsNotFoundErr(err) {
 		// Failed to GetInstance, however the Filestore instance may already be created.
 		// The error should be non-final.
-		return nil, status.Errorf(codes.Unavailable, "CreateVolume, failed to get instance: %v", err.Error())
+		return nil, common.NewTemporaryError(codes.Unavailable, fmt.Errorf("CreateVolume, failed to get instance: %w", err))
 	}
 
 	if filer != nil {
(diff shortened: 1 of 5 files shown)


riteshghorse:
CC: @epi-one

riteshghorse:
Friendly ping! @leiyiz @tyuchn

Pankaj (@eipi-one) will send another PR to update the metrics after this

pkg/csi_driver/multishare_stateful_controller.go
```diff
@@ -94,13 +94,13 @@ func (m *MultishareStatefulController) CreateVolume(ctx context.Context, req *cs
 	shareInfo, err := m.shareLister.ShareInfos(util.ManagedFilestoreCSINamespace).Get(pvName)
 	if err != nil {
 		if !errors.IsNotFound(err) {
-			return nil, status.Errorf(codes.Internal, "error getting shareInfo %q from informer: %s", pvName, err.Error())
+			return nil, status.Errorf(codes.Unavailable, "error getting shareInfo %q from informer: %s", pvName, err.Error())
```
leiyiz, line 97: any reason we don't do Temporary Error here?

pkg/csi_driver/controller.go
```diff
@@ -259,11 +262,11 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		if filer.State != "READY" {
 			msg := fmt.Sprintf("Volume %v not ready, current state: %v", name, filer.State)
 			klog.V(4).Infof(msg)
-			return nil, status.Error(codes.Internal, msg)
+			return nil, common.NewTemporaryError(codes.Unavailable, fmt.Errorf(msg))
```
leiyiz, line 265: in line#260 is DeadlineExceeded a temporary error? if not we should make that temporary too right

pkg/csi_driver/controller.go
```diff
@@ -241,7 +242,9 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 	filer, err := s.config.fileService.GetInstance(ctx, newFiler)
 	// No error is returned if the instance is not found during CreateVolume.
 	if err != nil && !file.IsNotFoundErr(err) {
-		return nil, file.StatusError(err)
+		// Failed to GetInstance, however the Filestore instance may already be created.
+		// The error should be non-final.
+		return nil, common.NewTemporaryError(codes.Unavailable, fmt.Errorf("CreateVolume, failed to get instance: %w", err))
```
leiyiz, line 247: we are at the CSI function level (CreateVolume) so we shouldn't directly return temporary error to the caller 

```
Do not return `TemporaryError` directly
// from CSI Spec API calls, as CSI Spec API calls MUST return a
// standard gRPC status.
```

pkg/csi_driver/multishare_stateful_controller.go
eipi-one: @leiyiz General question : What happens at the GKE/k8s end upon receiving such temporary errors? Will it retry to create the same instance? In order to track the ongoing creation the called needs the operation ID of previous request.

leiyiz: yeah it'll keep retrying on the same instance

pkg/csi_driver/controller.go
leiyiz: ^ please fix.

pkg/csi_driver/controller.go
```diff
@@ -183,7 +182,7 @@ func (s *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 
 		if err != nil {
 			klog.Errorf("CreateVolume returned an error %v, for request %+v", err, req)
-			return nil, err
+			return nil, file.StatusError(err)
```
leiyiz, line 185: we still need to return unavailable code to the caller of `CreateVolume`, now since we are at outermost, we need to unwrap the temporary error with the unavailable code.

see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1664 for detail. follow how the temporary error returned from the "internal" methods are propagated and see how it is unwrapped at the return point of `CreateVolume` call.

riteshghorse: the `file.StatusError` does the unwrapping of the error. I added a new check for temporary error [here](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/60c6e085ff3bfa0682d420b314a3ce435b57fc98/pkg/cloud_provider/file/file.go#L718) that handles it. 

I think what is missing from this PR is the handling the metrics for new Temporary error types. This will be done in separate PR.

riteshghorse: @leiyiz  WDYT?

pkg/csi_driver/multishare_controller.go
```diff
@@ -201,7 +202,7 @@ func (m *MultishareController) CreateVolume(ctx context.Context, req *csi.Create
 	// lock released. poll for op.
 	err = m.waitOnWorkflow(ctx, workflow)
 	if err != nil {
-		return nil, file.StatusError(fmt.Errorf("Create Volume failed, operation %q poll error: %w", workflow.opName, err))
+		return nil, common.NewTemporaryError(codes.Unavailable, fmt.Errorf("Create Volume failed, operation %q poll error: %w", workflow.opName, err))
```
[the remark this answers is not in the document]
riteshghorse: I think except this call at [#218](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/60c6e085ff3bfa0682d420b314a3ce435b57fc98/pkg/csi_driver/multishare_controller.go#L218)  that returns an internal error if there is no parent, everything else return a temporary error in that workflow. Let me know if understand it correctly.

PR #966, opened 2024-10-01
Add small band support in the zonal Filestore tier

### LINKED ISSUE #874: Add support for Filestore Zonal/Regional instances with 1-10Tb volume into Filestore CSI driver
Hello,

Accordingly to the official [documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/filestore-csi-driver):

> **Requirements**
>- To use the Filestore CSI driver, your clusters must use the proper GKE version number applicable to your service tier:
>   - Basic HDD with GKE version 1.21 or later
>   - Basic SSD with GKE version 1.21 or later
>   - Zonal (10 TiB to 100 TiB) with GKE version 1.27 or later
>   - Enterprise with GKE version 1.25 or later
>- To use the Filestore multishares capability, your clusters must use GKE version 1.25 or later.
>- The Filestore CSI driver is supported for clusters using Linux only; Windows Server nodes are not supported.
>- The minimum instance size for Filestore is at least 1 TiB. The minimum instance size depends on the Filestore service tier you - selected. To learn more, see [Service tiers](https://cloud.google.com/filestore/docs/service-tiers). 

Meanwhile, 10Tb volume is too expensive and overkill for many use cases. 

It'll be useful to have even basic/limited support of Filestore Zonal/Regional instances with 1-10Tb volume size in Filestore CSI driver with a clear information in the documentation.

riteshghorse:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup 

/kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adds support for small zonal band in the Filestore zonal tier.

Fixes #874

```release-note
Added support for Filestore small zonal band (1 Ti - 9.75 Ti)
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -53,8 +53,10 @@ const (
 	enterpriseTierMaxSize = 10 * util.Tb
 	highScaleTierMinSize  = 10 * util.Tb
 	highScaleTierMaxSize  = 100 * util.Tb
-	zonalTierMinSize      = 1 * util.Tb
-	zonalTierMaxSize      = 100 * util.Tb
+	zonalSmallTierMinSize = 1 * util.Tb
+	zonalSmallTierMaxSize = 9984 * util.Gb
+	zonalLargeTierMinSize = 10 * util.Tb
+	zonalLargeTierMaxSize = 100 * util.Tb
 	premiumTierMinSize    = 25 * util.Tb / 10
 	premiumTierMaxSize    = 639 * util.Tb / 10
 
@@ -104,6 +106,28 @@ const (
 	TagKeyClusterLocation          = "storage_gke_io_cluster_location"
 )
 
+// Parameters to define capacity range for Filestore tiers
+var (
+	defaultRange    = capacityRangeForTier{min: defaultTierMinSize, max: defaultTierMaxSize}
+	enterpriseRange = capacityRangeForTier{min: enterpriseTierMinSize, max: enterpriseTierMaxSize}
+	highScaleRange  = capacityRangeForTier{min: highScaleTierMinSize, max: highScaleTierMaxSize}
+	premiumRange    = capacityRangeForTier{min: premiumTierMinSize, max: premiumTierMaxSize}
+	zonalSmallRange = capacityRangeForTier{min: zonalSmallTierMinSize, max: zonalSmallTierMaxSize}
+	zonalLargeRange = capacityRangeForTier{min: zonalLargeTierMinSize, max: zonalLargeTierMaxSize}
+)
+
+// tierToCapacityRange maps tier names to their corresponding capacity ranges
+var tierToCapacityRange map[string]capacityRangeForTier = map[string]capacityRangeForTier{
+	defaultTier:    defaultRange,
+	enterpriseTier: enterpriseRange,
+	highScaleTier:  highScaleRange,
+	zonalTier:      zonalSmallRange,
+	premiumTier:    premiumRange,
+	basicSSDTier:   premiumRange, //these two are aliases
+	basicHDDTier:   defaultRange, //these two are aliases
+}
+
+// capacityRangeForTier represents minimum and maximum capacity values for a tier
 type capacityRangeForTier struct {
 	min int64
 	max int64
@@ -489,11 +513,11 @@ func getTierFromParams(params map[string]string) string {
 }
 
 // validator function to check for invalid capacity size requests
-func invalidCapacityRange(capRange *csi.CapacityRange, tier string, validRange *capacityRangeForTier) error {
+func invalidCapacityRange(requestedCapRange *csi.CapacityRange, tier string, validRange *capacityRangeForTier) error {
 
-	requiredCap := capRange.GetRequiredBytes()
+	requiredCap := requestedCapRange.GetRequiredBytes()
 	requireSet := requiredCap > 0
-	limitCap := capRange.GetLimitBytes()
+	limitCap := requestedCapRange.GetLimitBytes()
 	limitSet := limitCap > 0
 
 	if limitSet && requireSet && limitCap < requiredCap {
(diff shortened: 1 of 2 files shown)


riteshghorse:
R: @leiyiz @eipi-one

pkg/csi_driver/controller.go
```diff
@@ -524,34 +548,24 @@ func invalidCapacityRange(capRange *csi.CapacityRange, tier string, validRange *
...
-		premiumTier:    premiumRange,
-		basicSSDTier:   premiumRange, //these two are aliases
-		basicHDDTier:   defaultRange, //these two are aliases
-	}
-
+// provisionableCapacityForTier returns capacity range for tier
+func provisionableCapacityForTier(capRange *csi.CapacityRange, tier string) *capacityRangeForTier {
 	tier = strings.ToLower(tier)
-	validRange, ok := provisionableCapacityForTier[tier]
+	if tier == zonalTier && capRange != nil && capRange.GetRequiredBytes() > zonalLowTierMaxSize {
```
leiyiz, line 554: requestedCapRange maybe clearer name.

leiyiz, line 554: consider a case where first the zonal high tier is created, then a zonal small tier is created, wouldn't that result in requested Capacity not valid?

pkg/csi_driver/controller.go
```diff
@@ -723,7 +751,15 @@ func (s *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi.
...
 	if filer.State != "READY" {
-		return nil, fmt.Errorf("lolume %q is not yet ready, current state %q", volumeID, filer.State)
+		return nil, fmt.Errorf("Volume %q is not yet ready, current state %q", volumeID, filer.State)
+	}
+
+	// if the Filestore instance is in a small band of the tier then verify if the volume expansion
+	// request cross the band size.
+	invalidExpansion := invalidVolumeExpansionRequest(req.GetCapacityRange(), filer.Volume.SizeBytes, filer.Tier)
+	if invalidExpansion {
+		klog.Warningf("Volume expansion not supported beyond small %s band. Please create a new instance with higher storage capacity.", filer.Tier)
```
leiyiz, line 761: this log is repetitive of the above warningf logging, why not put this message inside the `status.Errorf`

pkg/csi_driver/controller.go
riteshghorse: Pardon, I didn't get your question. Could you please elaborate? I have tested this flow and that works. These will be different requests right with different PVC name?

pkg/csi_driver/controller.go
```diff
@@ -53,8 +53,10 @@ const (
 	enterpriseTierMaxSize = 10 * util.Tb
 	highScaleTierMinSize  = 10 * util.Tb
 	highScaleTierMaxSize  = 100 * util.Tb
-	zonalTierMinSize      = 1 * util.Tb
-	zonalTierMaxSize      = 100 * util.Tb
+	zonalLowTierMinSize   = 1 * util.Tb
```
eipi-one, line 56: I would recommend to use the same terminology of 'small' and 'large' instead of 'low' and 'high' to make it easy to read and understand.

pkg/csi_driver/controller.go
eipi-one: +1

eipi-one: This code appears a little buggy to me. We are updating the map with new entry every time and and immediately accessing it.

In that case let's not even create that entry in the map and evaluate the range on the fly depending on the zonal tier and band without caching it.

@leiyiz Is the CSI driver implementation thread safe in general? Can there be multiple create volume requests to the same driver? In that case this piece of code might not work as expected.

pkg/csi_driver/controller.go
riteshghorse: Good suggestion. Done.

pkg/csi_driver/controller.go
riteshghorse: yeah I think actually writing the map here is problematic. Fixed this by using the map for read only.

riteshghorse: > consider a case where first the zonal high tier is created, then a zonal small tier is created, wouldn't that result in requested Capacity not valid?

I see what you mean. Updated the code to not overwrite the map.

riteshghorse (commented):
Thanks for the comments. Updated the code. PTAL

riteshghorse:
@leiyiz @tyuchn this is good for another look, thanks!

pkg/csi_driver/controller.go
```diff
@@ -549,12 +549,12 @@ func invalidCapacityRange(capRange *csi.CapacityRange, tier string, validRange *
...
 // provisionableCapacityForTier returns capacity range for tier
-func provisionableCapacityForTier(capRange *csi.CapacityRange, tier string) *capacityRangeForTier {
+func provisionableCapacityForTier(requestedCapRange *csi.CapacityRange, tier string) *capacityRangeForTier {
 	tier = strings.ToLower(tier)
-	if tier == zonalTier && capRange != nil && capRange.GetRequiredBytes() > zonalLowTierMaxSize {
+	if tier == zonalTier && requestedCapRange != nil && requestedCapRange.GetRequiredBytes() > zonalSmallTierMaxSize {
 		// keep this check simple since the capacity bounds are checked thoroughly in the
 		// invalidCapacityRange() later.
-		tierToCapacityRange[zonalTier] = capacityRangeForTier{min: zonalHighTierMinSize, max: zonalHighTierMaxSize}
+		return &capacityRangeForTier{min: zonalLargeTierMinSize, max: zonalLargeTierMaxSize}
```
leiyiz, line 557: I think you can directly return `&zonalLargeRange` here right?

deploy/kubernetes/overlays/dev/controller_always_pull.yaml
```diff
@@ -6,6 +6,8 @@ spec:
   template:
     spec:
       containers:
-        - name: gcp-filestore-driver
-          imagePullPolicy: Always
+      - name: gcp-filestore-driver
+        imagePullPolicy: Always
+      imagePullSecrets:
+      - name: gcr-json-key
```
leiyiz, line 12: is the `imagePullSecrets` necessary?

deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -10,8 +10,7 @@ patchesStrategicMerge:
...
 namespace: gcp-filestore-csi-driver
 
-# To change the dev image, add something like the following.
-#images:
-#- name: gcr.io/gke-release/gcp-filestore-csi-driver
-#  newName: <your-gcp-project-image-url>
-#  newTag: <imagetag>
+images:
+- name: registry.k8s.io/cloud-provider-gcp/gcp-filestore-csi-driver
+  newName: gcr.io/riteshghorse-consumer/gcp-filestore-csi-driver
```
leiyiz, line 15: you probably don't want to include changes in this file.

leiyiz (commented):
after fix can you also squash the commits to 1 commit? that's customary for k8s sigs PR before submission.

deploy/kubernetes/overlays/dev/kustomization.yaml
riteshghorse: yeah looks like I accidentally committed these files yaml files

leiyiz (commented):
/lgtm
/approve

PR #983, opened 2024-11-01
Enable config map update retry mechanism during NodeStageVolume

siyanshen:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR implements a retry mechanism in NodeStageVolume- Instead of failing the NodeStageVolume operation after a single attempt to update the ConfigMap, a retry logic should be introduced. This would ensure that the operation continues to attempt updates until it succeeds, reducing the chance of unnecessary failures.

Why we need this? Before this change, there is a case where the CSI driver attempts to mount multiple Filestore-based volumes. During this process, the CSI driver stores lock info in a ConfigMap, which tracks the mounted Filestore information for each node. However, this can lead to conflicts when multiple NodeStageVolume calls attempt to update the ConfigMap simultaneously. In such cases, the update may fail because the key-value pairs have already been modified by another process. The resulting error message may look like this:
```
rpc error: code = Internal desc = failed to store lock info after NodeStageVolume succeeded on volume xxx to path xxx: Operation cannot be fulfilled on configmaps "fscsi-gke-xxx": the object has been modified; please apply your changes to the latest version and try again
```

This is not an unrecoverable error, as the operation will self-resolve during a subsequent NodeStageVolume call and eventually succeed within milliseconds.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/status"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	mount "k8s.io/mount-utils"
 	"sigs.k8s.io/gcp-filestore-csi-driver/pkg/cloud_provider/metadata"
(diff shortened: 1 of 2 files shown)


pkg/csi_driver/node.go
```diff
@@ -496,58 +497,56 @@ func getFSStat(path string) (available, capacity, used, inodesFree, inodes, inod
...
+			s.lockReleaseController.RecordKubeAPIMetrics(err, metrics.ConfigMapResourceType, metrics.CreateOpType, metrics.NodeStageOpSource, duration)
+			if err != nil {
+				klog.Errorf("NodeStageVolume failed to create configmap %+v with data %s for volume %s: %v", klog.KObj(cm), data, volumeID, err)
+				return err
+			}
+			klog.Infof("NodeStageVolume successfully created configmap %+v with data %v for volume %s", klog.KObj(cm), cm.Data, volumeID)
+			return nil
+		}
+
+		updateErr := s.lockReleaseController.UpdateConfigMapWithKeyValue(ctx, cm, lockInfoKey, filestoreIP)
```
tyuchn, line 546: I think this is the only place (UPDATE configmap) we need to have a retry mechanism, rather than the entire node stage volume function.

PR #999, opened 2024-11-07
Add lock release controller v2 feature flag

siyanshen:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds a feature flag for filestore lock release controller v2. This change ensures that v1 lock release controller has no regression, and v2 controller will only take effect when the flag is explicitly set as true in a new component release.

Fixes #

*Q: Why can't we reuse the existing feature-lock-release flag by turning it off during v2's component release?*
A: We can’t just turn off the feature flag because nodes need it to modify config map during volume staging/unstaging etc. That is why we need a v2 flag.
*Q: Where to put the v2 flag?*
A: `cmd/main.go` - We only need it where we switch between the logic of v1 and v2 (`cmd/main.go` & `gcfs_driver.go`). There’s no point of adding the feature flag in `cmd/lockrelease/main.go`. Because, to turn that part off is essentially to shut down the whole v2 container. At that point we can just roll back the component.

**Testing set up**
1. in  `deploy/kubernetes/base/controller/controller.yaml` and `deploy/kubernetes/base/node_linux/node.yaml`, 

  a. To test v2, add this following arg to gcp-filestore-driver container
```
- "--feature-lock-release=true"
- "--feature-lock-release-standalone=true"
```

  b. To verify v1 has no regression, disable the lock_release_controller, and add this following arg to gcp-filestore-driver container

```
- "--feature-lock-release=true"
```

2. override `gcp-filestore-csi-driver` and `filestore-lockrelease-controller` images with your testing images.
3. Run the following command
```
PROJECT=<project-name> GCFS_SA_DIR=<DIR>  ./deploy/project_setup.sh

# If need to rebuild csi driver 
GCP_FS_CSI_STAGING_VERSION=<tag> GCP_FS_CSI_STAGING_IMAGE=<target-image> make build-image-and-push

# If need to rebuild lock release controller
PROJECT=<project-name> make lockrelease-image

PROJECT=<project-name> DEPLOY_VERSION=lockrelease ./deploy/kubernetes/cluster_setup.sh
```

**Test cases covered** 
Following manual testing cases all passed.
1. v2 flag not set (default false), v2 lock release controller yaml nonexistent. This mimics the scenario before the component release containing our v2 controller rolls out.
_Expected Behavior_: v2 controller not created; v1 controller WAI.
2. v2 flag set as true. This mimics the scenario after our new component rolls out.
_Expected Behavior_: v1 does not compete against v2 for leader election; v1 does not try to do lock release at all.

```release-note
NONE
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -54,7 +54,9 @@ var (
 
 	// Feature lock release specific parameters, only take effect when feature-lock-release is set to true.
 	featureLockRelease = flag.Bool("feature-lock-release", false, "if set to true, the node driver will support Filestore lock release.")
-
+	// featureLockRelease must be set as true when featureLockReleaseV2 is true. V2 implementation will override part of v1 implementation when true.
+	featureLockReleaseV2  = flag.Bool("feature-lock-release-v2", false, "if set to true, the node driver will not support v1 Filestore lock release.")
+	lockReleaseSyncPeriod = flag.Duration("lock-release-sync-period", 60*time.Second, "Duration, in seconds, the sync period of the lock release controller. Defaults to 60 seconds.")
 	// Feature configurable shares per Filestore instance specific parameters.
 	featureMaxSharePerInstance = flag.Bool("feature-max-shares-per-instance", false, "If this feature flag is enabled, allows the user to configure max shares packed per Filestore instance")
 	descOverrideMaxShareCount  = flag.String("desc-override-max-shares-per-instance", "", "If non-empty, the filestore instance description override is used to configure max share count per instance. This flag is ignored if 'feature-max-shares-per-instance' flag is false. Both 'desc-override-max-shares-per-instance' and 'desc-override-min-shares-size-gb' must be provided. 'ecfsDescription' is ignored, if this flag is provided.")
@@ -165,11 +167,13 @@ func main() {
 
 	featureOptions := &driver.GCFSDriverFeatureOptions{
 		FeatureLockRelease: &driver.FeatureLockRelease{
-			Enabled: *featureLockRelease,
+			Enabled:    *featureLockRelease,
+			Standalone: *featureLockReleaseV2,
 			Config: &lockrelease.LockReleaseControllerConfig{
 				LeaseDuration:  *leaderElectionLeaseDuration,
 				RenewDeadline:  *leaderElectionRenewDeadline,
 				RetryPeriod:    *leaderElectionRetryPeriod,
+				SyncPeriod:     *lockReleaseSyncPeriod,
 				MetricEndpoint: *httpEndpoint,
 				MetricPath:     *metricsPath,
 			},
--- a/deploy/kubernetes/overlays/lockrelease/configmap_rbac.yaml
+++ b/deploy/kubernetes/overlays/lockrelease/configmap_rbac.yaml
@@ -31,7 +31,7 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["configmaps"]
-  verbs: ["get", "update", "create"]
+  verbs: ["get", "list", "update", "create"]
 
 ---
 
@@ -76,6 +76,21 @@ roleRef:
   name: filestorecsi-node-leaderelection-role
   apiGroup: rbac.authorization.k8s.io
 
+---
+
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: filestorecsi-builtin-controller-leaderelection-binding
+  namespace: gke-managed-filestorecsi
+subjects:
+- kind: ServiceAccount
+  name: gcp-filestore-csi-node-sa
+  namespace: gcp-filestore-csi-driver
+roleRef:
+  kind: Role
+  name: filestorecsi-node-leaderelection-role
+  apiGroup: rbac.authorization.k8s.io
 ---
 
 kind: RoleBinding
--- a/pkg/csi_driver/gcfs_driver.go
+++ b/pkg/csi_driver/gcfs_driver.go
@@ -131,8 +131,9 @@ type FeatureStateful struct {
 }
 
 type FeatureLockRelease struct {
-	Enabled bool
-	Config  *lockrelease.LockReleaseControllerConfig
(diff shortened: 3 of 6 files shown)


deploy/kubernetes/overlays/lockrelease/configmap_rbac.yaml
```diff
@@ -71,6 +71,9 @@ subjects:
 - kind: ServiceAccount
   name: filestore-lockrelease-controller-sa
   namespace: gcp-filestore-csi-driver
+- kind: ServiceAccount
```
saikat-royc, line 74: It's best to create a new role binding for the lock release controller `filestorecsi-lockrelease-controller-role-binding` and not couple the two

pkg/csi_driver/gcfs_driver.go
```diff
@@ -328,6 +329,10 @@ func (driver *GCFSDriver) Run(endpoint string) {
 	// Start the nonblocking GRPC.
 	s := NewNonBlockingGRPCServer()
 	s.Start(endpoint, driver.ids, driver.cs, driver.ns)
+	if driver.config.RunNode && driver.config.FeatureOptions.FeatureLockRelease.Enabled && !driver.config.FeatureOptions.FeatureLockRelease.V2 {
```
tyuchn, line 332: I don't think `v2` is a suitable name, as it might be too vague for others who lack context about the lock release functionality. How about naming the flag `builtin-lock-release-controller`? We could then structure the conditional like this:

```go
if driver.config.RunNode && driver.config.FeatureOptions.FeatureLockRelease.Enabled && driver.config.FeatureOptions.FeatureLockRelease.builtinLockReleaseController {
    // Start the lock release controller on the node driver.
    driver.ns.(*nodeServer).lockReleaseController.Run(context.Background())
}
```

This makes the purpose of the flag clearer to anyone reading the code.

cmd/main.go
```diff
@@ -53,8 +53,9 @@ var (
 	resourceTagsStr                 = flag.String("resource-tags", "", "Resource tags to attach to each volume created. It is a comma separated list of tags of the form '<parentID_1>/<tagKey_1>/<tagValue_1>...<parentID_N>/<tagKey_N>/<tagValue_N>' where, parentID is the ID of Organization or Project resource where tag key and value resources exist, tagKey is the shortName of the tag key resource, tagValue is the shortName of the tag value resource. See https://cloud.google.com/resource-manager/docs/tags/tags-creating-and-managing for more details.")
 
 	// Feature lock release specific parameters, only take effect when feature-lock-release is set to true.
-	featureLockRelease = flag.Bool("feature-lock-release", false, "if set to true, the node driver will support Filestore lock release.")
-
+	featureLockRelease    = flag.Bool("feature-lock-release", false, "if set to true, the node driver will support Filestore lock release.")
+	featureLockReleaseV2  = flag.Bool("feature-lock-release-v2", false, "if set to true, the node driver will not support Filestore lock release.")
```
saikat-royc, line 57: "... not support Filestore v1 lock release"

pkg/csi_driver/gcfs_driver.go
saikat-royc: +1 on a more intuitive name other than v2. But @tyuchn can you please elaborate your thought on what does "builtin" mean

pkg/csi_driver/gcfs_driver.go
```diff
@@ -132,6 +132,7 @@ type FeatureStateful struct {
 
 type FeatureLockRelease struct {
 	Enabled bool
```
saikat-royc, line 134: also if we adding variable for v2 we can as well change this variable `Enabled` to make it consistent

pkg/csi_driver/gcfs_driver.go
tyuchn: I meant that the lock release controller runs as part of the CSI driver. Perhaps builtin isn’t the best term, but my point is to use a more descriptive flag to control whether the lock release controller runs in standalone mode or as part of the CSI driver.

saikat-royc: how about we change the names to 

type FeatureLockRelease struct {
    builtIn bool # <description>
    standAlone bool # <description>
}

cmd/main.go
saikat-royc: also looking at the implementation if both set to true, the v2 implementation take precedence. we should clearly call that out in the flag comment

pkg/csi_driver/gcfs_driver.go
siyanshen: I agree with the Standalone idea, but I will keep "Enabled". The flag is also used to make sure config maps are modified when volumes are staging/unstaging. This behavior does not change when we use the standalone controller. 

Think of the v1 feature flag as the light switch, and v2 feature flag as the switch of the light bulb's voice control function...

pkg/csi_driver/gcfs_driver.go
siyanshen: discussed in comment above

cmd/main.go
siyanshen: done

cmd/main.go
```diff
@@ -54,7 +54,9 @@ var (
 
 	// Feature lock release specific parameters, only take effect when feature-lock-release is set to true.
 	featureLockRelease = flag.Bool("feature-lock-release", false, "if set to true, the node driver will support Filestore lock release.")
-
+	// featureLockRelease must be set as true when featureLockReleaseV2 is true. V2 implementation will override part of v1 implementation when true.
+	featureLockReleaseV2  = flag.Bool("feature-lock-release-v2", false, "if set to true, the node driver will not support v1 Filestore lock release.")
```
tyuchn, line 58: Since we’re the only ones familiar with the differences between v1 and v2, should we consider naming it `standalone-lock-release-controller` to make it more intuitive for others?

tyuchn (changes requested):
/lgtm

PR #1073, opened 2025-02-25
Round up max capacity if limit is not set

### LINKED ISSUE #1059: Allow Filestore CSI driver to provision a max capacity volume when the request capacity is beyond limit
The Filestore CSI driver returns an error when the requested capacity is beyond the Filestore tier limit https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/2c489f3493f1f0ed4ad9ed53c7a33dac030b760a/pkg/csi_driver/controller.go#L232. 

This creates a retry-loop to provision this resource at some point. Instead of doing this, it is better for users to provision a max capacity instance with a warning that `requested capacity is more than the tier capacity limit, provisioning a max tier capacity instance`

riteshghorse:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #1059

```release-note
Provision a Filestore instance with max capacity if the required capacity is more than the max allowed and the LimitBytes are not set.
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -529,13 +529,15 @@ func invalidCapacityRange(requestedCapRange *csi.CapacityRange, tier string, val
 	}
 
 	if requireSet {
-		if requiredCap > validRange.max {
-			return fmt.Errorf("request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s", float64(requiredCap)/util.Tb, float64(validRange.max)/util.Tb, tier)
+		if !limitSet && requiredCap > validRange.max {
+			klog.Warningf("Request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s, but no upper bound was specified. Rounding off capacity request to %vTiB for tier %s", float64(requiredCap)/util.Tb, float64(validRange.max)/util.Tb, tier, float64(validRange.max)/util.Tb, tier)
+		} else if requiredCap > validRange.max {
+			return fmt.Errorf("Request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s", float64(requiredCap)/util.Tb, float64(validRange.max)/util.Tb, tier)
 		}
 
 		if !limitSet && requiredCap < validRange.min {
 			// Avoid surprising users by provisioning more than Requested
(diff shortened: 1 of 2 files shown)


riteshghorse:
R: @eipi-one @leiyiz

pkg/csi_driver/controller.go
```diff
@@ -529,7 +529,9 @@ func invalidCapacityRange(requestedCapRange *csi.CapacityRange, tier string, val
 	}
 
 	if requireSet {
-		if requiredCap > validRange.max {
+		if !limitSet && requiredCap > validRange.max {
+			klog.Warningf("request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s, but no upper bound was specified. Rounding up capacity request to %vTiB for tier %s", float64(requiredCap)/util.Tb, float64(validRange.max)/util.Tb, tier, float64(validRange.max)/util.Tb, tier)
```
eipi-one, line 533: Nit: change to "off".

pkg/csi_driver/controller.go
```diff
@@ -529,7 +529,9 @@ func invalidCapacityRange(requestedCapRange *csi.CapacityRange, tier string, val
 	}
 
 	if requireSet {
-		if requiredCap > validRange.max {
+		if !limitSet && requiredCap > validRange.max {
```
eipi-one, line 532: I remember we decided to created large zonal in case the capacity requested is more than max for small zonal. Will this in any way interfere with that?

riteshghorse: No, if capacity is more than the small zonal, we assign a large zonal band in a function prior to this - https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/518f22d7e0edae391e321e312a870daa70856263/pkg/csi_driver/controller.go#L558

So this won't interfere.

pkg/csi_driver/controller.go
```diff
@@ -529,7 +529,9 @@ func invalidCapacityRange(requestedCapRange *csi.CapacityRange, tier string, val
...
+			klog.Warningf("request bytes %vTiB is more than maximum instance size bytes %vTiB for tier %s, but no upper bound was specified. Rounding off capacity request to %vTiB for tier %s", float64(requiredCap)/util.Tb, float64(validRange.max)/util.Tb, tier, float64(validRange.max)/util.Tb, tier)
```
tyuchn, line 533: nit: we should capitalize in warning

PR #1109, opened 2025-04-15
Set protocol field for Multishare instances

riteshghorse:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The multishare_autopilot test is failing because it couldn't find the right matching instance for the share because the protocol value is set in the request but in the found source instances.

Fixes #

```release-note
NONE
```


riteshghorse:
R: @tyuchn

PR #1314, opened 2026-06-08
[WIP] Write dummy file to skip gVisor emptyDir optimization since the direc…

amacaskill:
…tory isn't empty

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/pkg/csi_driver/node.go
+++ b/pkg/csi_driver/node.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -41,8 +42,8 @@ import (
 )
 
 const (
-	optionSmbUser     = "smbUser"
-	optionSmbPassword = "smbPassword"
+	optionSmbUser      = "smbUser"
+	optionSmbPassword  = "smbPassword"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


pkg/csi_driver/node.go
```diff
@@ -175,6 +176,21 @@ func (s *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 
 	klog.V(4).Infof("Successfully mounted %s on node %s", targetPath, s.driver.config.NodeName)
+
+	if value, ok := volumeContext[util.AttrLateBindSubdir]; ok && value != "" {
+		lateBindPath := filepath.Join(targetPath, value)
+		if err := os.MkdirAll(lateBindPath, 0750); err != nil {
+			klog.Errorf("Failed to create late-bind subdirectory %s: %v", lateBindPath, err)
+		} else {
+			dirtyFilePath := filepath.Join(lateBindPath, ".gke-late-bind-dirty")
```
amacaskill, line 185: The subdirectory doesn't have to have a file in it, since the volume mounted in NPV isn't an emptyDir, it isn't optimized away. It just needs to exist at mount, since we can't add the subDir after the fact and we can't replace the top dir. The bindmount is going to be linked to the subDir at claim time, but subDir needs to be created at pod creation time, before gVisor finishes setting up its mount tables..

PR #1349, opened 2026-08-05
Add official go client for sharepool APIs

ayushi-gupta:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

* Use official go client for new sharepool APIs instead of using https calls as the official go clients now expose the filestore sharepool APIs.
* Some checks related to sharepool unsupported operations are also added.
* An example to use sharepool is also added.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/csi_driver/controller.go
+++ b/pkg/csi_driver/controller.go
@@ -557,6 +557,10 @@ func (s *controllerServer) ControllerModifyVolume(ctx context.Context, req *csi.
 		return nil, status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
 	}
 
+	if isSharePoolVolumeID(volumeID) {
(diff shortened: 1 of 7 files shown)


pkg/csi_driver/controller.go
```diff
@@ -557,6 +557,10 @@ func (s *controllerServer) ControllerModifyVolume(ctx context.Context, req *csi.
 		return nil, status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
 	}
 
+	if isSharePoolVolumeID(volumeID) {
```
savirg, line 560: Please add unit tests for these changes

savirg (changes requested):

ayushi-gupta:
The sharepool APIs are not going to be supported in the newer versions.

PR #1360, opened 2026-08-24
Add support for FiFA VolumePool volume provisioning

arokade-dev:
**What type of PR is this?**

/kind feature

This PR introduces support for **FiFA (Filestore For Agent) VolumePools** in the GCP Filestore CSI Driver. 

Key enhancements:
- **Sub-Second Dynamic Provisioning**: Enables GKE workloads (especially AI/ML, batch processing, and serverless jobs) to claim micro-volumes in milliseconds from pre-warmed, multi-share Filestore Enterprise pools.
- **AIP Standardized VolumePool Integration**: Integrates `CreateVolume` and `DeleteVolume` calls with the `volumePools/{pool}/volumes` REST/gRPC endpoints.
- **Dynamic Mount Point Resolution**: Resolves and mounts the pre-allocated NFS share name (`mountName`) and internal IP (`ipAddress`) returned by the VolumePool backend.

Fixes #

- **StorageClass Parameter**: Workloads enable this by specifying `volume-pool: "projects/<project>/locations/<location>/volumePools/<pool-name-or-uuid>"` in their Kubernetes `StorageClass`.
- **Validation & Testing**: Verified end-to-end on GKE with dynamic PVC creation, sub-second volume binding, and successful NFS read/write mounting inside test pods.

```release-note
Add support for the `volume-pool` StorageClass parameter to enable fast dynamic provisioning from FiFA VolumePools.
```


deploy/kubernetes/overlays/volumepool/controller_patch.yaml
```diff
@@ -31,4 +32,5 @@ spec:
         - "--endpoint=unix:/csi/csi.sock"
         - "--nodeid=$(KUBE_NODE_NAME)"
         - "--controller=true"
-        - "--feature-share-pools=true"
+        - "--feature-volume-pools=true"
+        - "--filestore-service-endpoint=staging-file.sandbox.googleapis.com"
```
arokade-dev: Remove

deploy/kubernetes/overlays/volumepool/kustomization.yaml
```diff
@@ -9,6 +9,6 @@ namespace: gcp-filestore-csi-driver
 
 images:
 - name: registry.k8s.io/cloud-provider-gcp/gcp-filestore-csi-driver
-  newName: gcr.io/gke-release/gcp-filestore-csi-driver
-  digest: sha256:9bde4b2e81e914c10681298442fbe39bb5be597d82fb91d1f11589ecc2284af0
+  newName: gcr.io/arokade-consumer/gcp-filestore-csi-driver
```
arokade-dev, line 13: replace it with gke-release repo

ayushi-gupta: Just add a TODO.

pkg/cloud_provider/file/file.go
```diff
@@ -199,9 +212,10 @@ type Service interface {
 	GetOp(ctx context.Context, op string) (*filev1beta1multishare.Operation, error)
 	IsOpDone(op *filev1beta1multishare.Operation) (bool, error)
 	ListOps(ctx context.Context, resource *ListFilter) ([]*filev1beta1multishare.Operation, error)
-	// Share pool ops
-	AcquireShare(ctx context.Context, parentPool string, requestID string, capacityGb int64) (*PoolShare, error)
-	ReleaseShare(ctx context.Context, poolName string, ipAddress string, shareID string) error
+	// Volume pool ops
```
ayushi-gupta, line 214: Just CreateVolume, DeleteVolume, GetVolume?

pkg/cloud_provider/file/file.go
```diff
@@ -1486,80 +1500,112 @@ const (
...
+		return googleapi.CheckResponse(resp)
+	}
+	return nil
 }
 
-func (manager *gcfsServiceManager) ReleaseShare(ctx context.Context, poolName string, ipAddress string, shareID string) error {
-	reqBody := &ReleaseShareRequest{
-		IpAddress: ipAddress,
-		ShareId:   shareID,
+func (manager *gcfsServiceManager) GetVolumePoolVolume(ctx context.Context, volumeURI string) (*PoolVolume, error) {
```
ayushi-gupta: Are we using GetVolume anywhere? If not, I would recommend to removing it and we integrate it with proper go SDK when required.

pkg/cloud_provider/file/file.go
```diff
@@ -1486,80 +1500,112 @@ const (
...
 
-	return &PoolShare{
-		IpAddress: respBody.IpAddress,
-		ShareId:   respBody.ShareId,
-	}, nil
+	resp, err := manager.httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send DeleteVolume request: %w", err)
+	}
+	defer resp.Body.Close()
```
ayushi-gupta: This is all common code in create and delete, would it be better to wrap it in a separate method and keep these clean to avoid duplication?

arokade-dev:
Hi @ayushi-gupta @msau42  Could you please review this PR when you have some time? Please let me know if you have any remaining feedback or if this looks good to go.

arokade-dev:
> Hi @ayushi-gupta @msau42 Could you please review this PR when you have some time? Please let me know if you have any remaining feedback or if this looks good to go.

@ayushi-gupta @msau42 Gentle Reminder!

pkg/cloud_provider/file/file.go
ayushi-gupta: We don't want to use Share term at all, Share is internal and physical representation of volume.

pkg/cloud_provider/file/file.go
```diff
@@ -1486,80 +1498,82 @@ const (
...
+		jsonBytes, err := json.Marshal(reqBody)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal request body: %w", err)
+		}
+		req, err := http.NewRequestWithContext(ctx, method, url, bytes.NewBuffer(jsonBytes))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create http request: %w", err)
+		}
+		req.Header.Set("Content-Type", "application/json")
+		return manager.httpClient.Do(req)
```
ayushi-gupta, line 1518: Do we need timeout/retries, since the k8 sidecard retry mechanisms don't automatically apply to substrate control plane, we need to confirm if we need to add anything in the csi layer.

ayushi-gupta (changes requested):

pkg/cloud_provider/file/file.go
ayushi-gupta: Also, acquire and release are no more supported, we should use standard methods Create/Delete.

pkg/cloud_provider/file/file.go
arokade-dev: Good question! I think we should actually avoid internal retries here and stick to the standard k8s CSI pattern: the driver should be "fail-fast" and the caller (ate-api-server) should handle the retry/backoff loop, exactly like K8s sidecar reconcilers do.

@ayushi-gupta please let me know your thought on it.

deploy/kubernetes/overlays/volumepool/kustomization.yaml
```diff
@@ -9,6 +9,7 @@ namespace: gcp-filestore-csi-driver
 
 images:
 - name: registry.k8s.io/cloud-provider-gcp/gcp-filestore-csi-driver
-  newName: gcr.io/gke-release/gcp-filestore-csi-driver
-  digest: sha256:9bde4b2e81e914c10681298442fbe39bb5be597d82fb91d1f11589ecc2284af0
+  # TODO: replace it with gke-release repo
+  newName: gcr.io/arokade-consumer/gcp-filestore-csi-driver
```
savirg, line 13: why are we using personal project image here?

examples/kubernetes/volumepool/storageclass.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: staging-volume-pool-sc
```
savirg: can we change it to volume-pool-sc to keep it consistent with the other resource names?

pkg/cloud_provider/file/file.go
```diff
@@ -1486,80 +1498,82 @@ const (
...
+		return nil, fmt.Errorf("failed to create http request: %w", err)
 	}
-	req.Header.Set("Content-Type", "application/json")
+	return manager.httpClient.Do(req)
+}
 
-	resp, err := manager.httpClient.Do(req)
-	if err != nil {
-		return fmt.Errorf("failed to send request for %s: %w", apiMethod, err)
+func (manager *gcfsServiceManager) AcquireVolumePoolShare(ctx context.Context, parentPool string, volumeID string) (*PoolVolume, error) {
```
savirg: Should we use CreateVolume here instead of AcquireVolumePoolShare?

savirg (changes requested):

pkg/cloud_provider/file/file.go
ayushi-gupta: I believe this should be the case for substrate control plane and the semantics are not changed from standard k8 CSI pattern, but since we're the first to integrate, I think we should confirm this with @msau42 that this indeed is the guidance for sustrate too (which I think should be).

arokade-dev: Yes, definitely! I completely agree we should loop in @msau42 to confirm this guidance.

Additionally, since multiple different CSI drivers will eventually integrate with the ATE system, enforcing the orchestration (timeouts and retry loops) centrally within the Substrate control plane is by far the most robust pattern. Establishing this standard now ensures that all future CSI plugins can easily integrate with Substrate.

examples/kubernetes/volumepool/storageclass.yaml
arokade-dev: Yes, I've updated the name to volume-pool-sc to keep it consistent

pkg/cloud_provider/file/file.go
arokade-dev: I would prefer to keep it as AcquireVolumePoolShare. We aren't creating a new instance from scratch here also we're just acquiring a share from a pre-provisioned pool. I feel this naming more accurately reflects the VolumePool API semantics.

deploy/kubernetes/overlays/volumepool/kustomization.yaml
arokade-dev: Good catch! As noted in the TODO, an official public release image supporting VolumePools isn't published just yet. I am using this temporarily and will swap it to the official image in my subsequent PR ([#1366)](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/1366).

pkg/cloud_provider/file/file.go
ayushi-gupta: I understand, but CSI driver doesn't need to know if the Volume was acquired or created. The original name of the API was acquire, hence it was integrated this way. The official API also have been renamed the to createVolume and DeleteVolme, we should use the same in CSI driver to keep the implementation clean and understandable. Also, as stated earlier, pool share is no more a concept for customer, volume is the only known construct from customer standpoint and the only logical resource our public API support. I would rather rename this to CreateVolumePoolVolume.

pkg/cloud_provider/file/file.go
msau42: substrate does not continuously reconcile as it is a grpc service. However the user is expected to retry the call. So I would agree with no internal retries

PR #1370, opened 2026-09-02
Add defensive rpc.statd verification, liveness probes, and writable /run mounts

arokade-dev:
**What type of PR is this?**
>
/kind bug

This heavily stabilizes the `nfs-services` container against a known race condition where crashed or zombie `rpc.statd` daemons silently block successful NFS GCE Filestore volume mounting. 

Previously, if `statd` failed to launch or became orphaned by a network reset, the container's primary bash loop (`sleep infinity`) would remain fully active. Kubelet would blindly view the pod as "Running" forever despite NFS locking being completely crippled on the underlying node.

This PR introduces comprehensive diagnostic and self-healing infrastructure:
1. **Writable `/run` volume override (`emptyDir`)**: The K8s Node container executes with strict security profiles (`readOnlyRootFilesystem: true`). An `emptyDir` mount (`rpc-run-dir`) was added over `/run` to guarantee `statd` dynamically possesses write-permissions for resolving local PID states and lockfiles.
2. **Defensive `/run` Re-Mapping (`nfs_services_start.sh`)**: Because the `emptyDir` intentionally eclipses the pre-compiled Docker directories, we inject `mkdir -p` commands at boot to recreate `/run/sendsigs.omit.d` and `/run/rpc_pipefs` so the Debian SysVinit handlers (like `rpcbind`) boot up silently and cleanly. 
3. **Rigorous Verification Loop (`nfs_services_start.sh`)**: The script now actively queries `rpcinfo` for 5 retries. If port 100024 fails to bind, the script purposefully executes `/sbin/rpc.statd -F -d` in the foreground to stream critical debug crashes directly to Cloud Logging, before executing an intentional container exit (`exit 1`) for node restarting.
4. **Liveness & Readiness Probes (`node.yaml`)**: The Kubelet has been configured to actively ping both port `111` (rpcbind) and port `100024` (statd). If either mapping dies across 3 consecutive 15-second cycles, Kubernetes detects the zombie state and successfully destroys/refreshes the sidecar.

Fixes b/398437622

*Important manual mitigation for currently degraded clusters:*
Because the previous implementation lacked these probes and utilized `hostNetwork: true`, it's possible that degraded nodes currently have completely zombified `rpc.statd` processes permanently gripping port `100024` on the host OS itself. Because the host port is blocked, the newly rolled-out container will fail to bind it. 
To clear this state when rolling out this fix on severely degraded environments, cluster operators must manually execute `pkill rpc.statd` on the physical K8s node via SSH, or simply perform a physical rolling machine reboot of the impacted Linux node instances.

```release-note
Fixed a critical issue where zombie NFS `rpc.statd` processes would permanently block successful Filestore volume mounts. Added a writable `rpc-run-dir` mount for stable locking and active Kubelet liveness/readiness probes to natively detect and restart corrupted NFS services.
```

### DIFF
--- a/deploy/kubernetes/base/node_linux/node.yaml
+++ b/deploy/kubernetes/base/node_linux/node.yaml
@@ -63,6 +63,17 @@ spec:
         - name: nfs-services
           image: registry.k8s.io/cloud-provider-gcp/gcp-filestore-csi-driver
           command: ["/nfs_services_start.sh"]
+          volumeMounts:
+            - mountPath: /run
+              name: rpc-run-dir
+          livenessProbe:
+            exec:
+              command: ["rpcinfo", "-T", "udp", "127.0.0.1", "100024"]
+            initialDelaySeconds: 10
+            periodSeconds: 15
+          readinessProbe:
+            exec:
+              command: ["rpcinfo", "-p", "127.0.0.1"]
       volumes:
         - name: registration-dir
           hostPath:
@@ -76,3 +87,5 @@ spec:
           hostPath:
             path: /var/lib/kubelet/plugins/filestore.csi.storage.gke.io/
             type: DirectoryOrCreate
+        - name: rpc-run-dir
+          emptyDir: {}
--- a/deploy/kubernetes/manifests/node.yaml
+++ b/deploy/kubernetes/manifests/node.yaml
@@ -59,6 +59,17 @@ spec:
           # Modify this image for local development with master branch.
           image: gcr.io/gke-release/gcp-filestore-csi-driver:v0.2.0-gke.0
           command: ["/nfs_services_start.sh"]
+          volumeMounts:
+            - mountPath: /run
+              name: rpc-run-dir
+          livenessProbe:
+            exec:
+              command: ["rpcinfo", "-T", "udp", "127.0.0.1", "100024"]
+            initialDelaySeconds: 10
+            periodSeconds: 15
+          readinessProbe:
+            exec:
+              command: ["rpcinfo", "-p", "127.0.0.1"]
       volumes:
         - name: registration-dir
           hostPath:
@@ -72,3 +83,5 @@ spec:
           hostPath:
             path: /var/lib/kubelet/plugins/filestore.csi.storage.gke.io/
             type: DirectoryOrCreate
+        - name: rpc-run-dir
+          emptyDir: {}
--- a/deploy/kubernetes/nfs_services_start.sh
+++ b/deploy/kubernetes/nfs_services_start.sh
@@ -3,7 +3,11 @@
 set -o errexit
 trap "exit 0" TERM
 
-# 1. Start rpcbind safely (don't fail errexit if port 111 is already bound)
+# 1. Recreate paths shadowed by the Kubernetes emptyDir mount over /run
+mkdir -p /run/sendsigs.omit.d
+mkdir -p /run/rpc_pipefs
+
+# 2. Start rpcbind safely (don't fail errexit if port 111 is already bound)
 if ! rpcinfo -p 127.0.0.1 >/dev/null 2>&1; then
   service rpcbind start || echo "rpcbind start returned non-zero, continuing..."
 fi
--- a/deploy/kubernetes/nfs_services_start.sh
+++ b/deploy/kubernetes/nfs_services_start.sh
@@ -1,23 +1,50 @@
 #!/bin/sh
 
 set -o errexit
+trap "exit 0" TERM
 
-trap "{ exit 0; }" TERM
-
-service rpcbind start
-
-# If statd is already running, for example becuase of an existing nfs mount, we will fail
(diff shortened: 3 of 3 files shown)


deploy/kubernetes/base/node_linux/node.yaml
```diff
@@ -76,3 +87,5 @@ spec:
           hostPath:
             path: /var/lib/kubelet/plugins/filestore.csi.storage.gke.io/
             type: DirectoryOrCreate
+        - name: rpc-run-dir
```
tyuchn, line 90: During a DaemonSet rolling update, emptyDir gets wiped when the Pod is replaced. That erases /run/rpcbind warm-start state (so the new rpcbind loses the kernel lockd registration for existing NFSv3 mounts) and NSM lock state. Also, nfs-services doesn't actually have readOnlyRootFilesystem: true (that's on csi driver, which only tried to run start-statd when statd wasn't registered). Instead of an emptyDir on /run, could we use hostPath so state survives Pod restarts?

deploy/kubernetes/manifests/node.yaml
```diff
@@ -59,6 +59,17 @@ spec:
...
           image: gcr.io/gke-release/gcp-filestore-csi-driver:v0.2.0-gke.0
           command: ["/nfs_services_start.sh"]
+          volumeMounts:
+            - mountPath: /run
+              name: rpc-run-dir
+          livenessProbe:
+            exec:
+              command: ["rpcinfo", "-T", "udp", "127.0.0.1", "100024"]
+            initialDelaySeconds: 10
+            periodSeconds: 15
```
tyuchn, line 69: Can we drop readinessProbe and bump livenessProbe.periodSeconds to 30 (or 60)?

`rpcinfo -T udp 127.0.0.1 100024` already checks both rpcbind (port 111) and statd (100024), and Kubelet doesn't check Pod readiness before calling the CSI socket anyway. Since this DaemonSet runs on every node with only 10m CPU request on nfs-services, running two exec probes every 10–15s adds 10 runc exec calls/min per node across the cluster.

deploy/kubernetes/nfs_services_start.sh
```diff
@@ -1,23 +1,54 @@
 #!/bin/sh
 
 set -o errexit
+trap "exit 0" TERM
```
tyuchn, line 4: In /bin/sh, a TERM trap won't fire while waiting on a foreground sleep infinity child. On Pod termination, nfs-services will ignore SIGTERM, hang for the full 30s grace period while gcp-filestore-driver is already dead, and then get SIGKILLed before rpcbind -w can save its warm-start state.

Consider stopping the services in the trap and background sleep, e.g.
```
trap 'service nfs-common stop 2>/dev/null || true; service rpcbind stop 2>/dev/null || true; exit 0' TERM INT
...
sleep infinity &
wait $!
```

deploy/kubernetes/nfs_services_start.sh
```diff
@@ -1,23 +1,54 @@
...
+MAX_RETRIES=5
+COUNT=0
+until rpcinfo -T udp 127.0.0.1 100024 >/dev/null 2>&1; do
+  COUNT=$((COUNT + 1))
+  if [ "$COUNT" -ge "$MAX_RETRIES" ]; then
+    echo "--- DIAGNOSTIC FAILURE ---" >&2
+    echo "statd failed to register after $MAX_RETRIES attempts." >&2
+    echo "Attempting to launch statd in foreground for debug logs..." >&2
+    
+    # Kill any zombie background process that might be hanging
```
tyuchn, line 39: If /sbin/rpc.statd -F -d happens to succeed on this attempt (e.g., slow startup), -F will keep it running in the foreground forever with -d debug logging enabled, spamming stderr for the life of the container. Also note that pkill here only sees the container PID namespace, not the host.

tyuchn (approved):

arokade-dev:
Hi @tyuchn,

Thanks for the detailed review! I've updated the PR to address all of your feedback:

* **rpcbind state persistence:** Switched the /run volume from emptyDir to hostPath to ensure the rpcbind.state file is preserved across pod restarts, preventing any loss of NFS active locks during upgrades.
* **Probe overhead reduction:** Removed the redundant readinessProbe and increased the livenessProbe interval to 30 seconds to minimize exec probe overhead across the cluster.
* **Pod shutdown hang fix:** Updated sleep infinity to sleep infinity & wait $! in the entrypoint script. This ensures SIGTERM is no longer blocked and the trap function executes immediately, allowing a clean and graceful shutdown without hitting Kubelet's 30-second forced SIGKILL timeout.
* **Crash loop behavior:** Removed the blocking foreground fallback process (rpc.statd -F -d) and replaced it with an explicit exit 1. Now, if statd fails to register, the container will cleanly crash and trigger Kubernetes' standard CrashLoopBackOff recovery.


PTAL when you have a chance!

